PR #32, opened 2016-10-24
Slate product section

cshold:
- Move show/hide social sharing buttons to product template and featured product sections
- Change all product.js related selectors to classes and their uses to `$('.selector-name', this.$container)`
  - Allows multiple sections on the same page without adding `section.id` to their IDs
  - Best perf option for this type of selectors. [Source](http://stackoverflow.com/a/3177782/412106)
- `enableHistoryState` setting isn't required and defaults to false. Data attribute only needs to be set on the product template

Editor demo - https://carson-blocks.myshopify.com/admin/themes/132761987/editor
Home page with three featured products - http://carson-blocks.myshopify.com/
Product page with a few options - http://carson-blocks.myshopify.com/products/burton-dunmore-jacket-2016

@Shopify/themes-fed

### DIFF
--- a/src/config/settings_schema.json
+++ b/src/config/settings_schema.json
@@ -64,13 +64,7 @@
     "settings": [
       {
         "type": "header",
-        "content": "Sharing options"
-      },
-      {
-        "type": "checkbox",
-        "id": "social_sharing_products",
-        "label": "Enable sharing for products",
-        "default": true
+        "content": "Social sharing options"
       },
       {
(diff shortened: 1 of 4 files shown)


iamkeir:
I _think_ this allows for multiple product sections on a single page - but is there any guidance on how to implement this please? I'm not sure how to call/invoke the product JS functionality (variants + image select) for multiple instances.

Issue #83, opened 2017-01-18
dev environment points to live


NathanPJF:
*From conversation in https://github.com/Shopify/slate-tools/issues/39*

We are having the `development` environment point to the live theme by default.  The is where a `production` environment should point.  We should encourage pushing to a safe dev environment by default.

jmcmullen:
+1

NathanPJF:
@jmcmullen If you're finding your theme_id isn't being respected then I believe the issue you're experiencing is related to this one: https://github.com/Shopify/slate-cli/issues/115

NathanPJF:
Fixed in https://github.com/Shopify/slate/pull/84

PR #84, opened 2017-01-18
add production env

NathanPJF:
Fix for https://github.com/Shopify/slate/issues/83

Partial fix for https://github.com/Shopify/slate-tools/issues/39

@macdonaldr93 @cshold @t-kelly Any concerns about this?  Alternatively, I just remove "live" as the initial value.


NathanPJF:
Merging this as a band-aid fix for the issues reported in https://github.com/Shopify/slate-tools/issues/39.

I recognize this requires an extra step when testing, but it's worth it to avoid having devs accidentally wipe their live site.  You can always write the env flag if you don't want to go looking for the theme id: `slate deploy -e production`

Also, now the `config.yml` has an example of how to create multiple environments in your config file. 😎

Issue #81, opened 2017-01-12
[Docs] Explain /= require


NathanPJF:
## Documentation

Let's put a note in the docs [JavaScript helpers](https://shopify.github.io/slate/theme-scaffold/#javascript-helpers) doc `// =require` is a Gulp thing and where to find more info: https://www.npmjs.com/package/gulp-include

Issue #69, opened 2016-12-18
[Docs] add documentation on sections.js


NathanPJF:
In the [JS examples](https://shopify.github.io/slate/js-examples) docs, we need to include documentation on how to use this help to `register` new sections.

Once added, we also will need to add it to the sidebar.

PR #94, opened 2017-02-17
add docs about section events and registering

NathanPJF:
Fix https://github.com/Shopify/slate/issues/69
Fix https://github.com/Shopify/slate/issues/81


docs/js-examples/index.md
```diff
@@ -192,4 +192,60 @@ The theme-specific script of `product.js` has a number of methods that listen fo
...
+</div>
+```
+{% endraw %}
+
+In the `theme.js` file, you must import the Section specific JavaScript with the `// =require` helper, [more information here](https://www.npmjs.com/package/gulp-include), as it will contain your `constructor`. 
+
+```
+// =require sections/product.js
+
+$(document).ready(function() {
```
JesseBarnett24: if i wanted to use $(document).ready(function() in another smaller js file, is it still required to register the section?

PR #101, opened 2017-02-27
Add cart cookies disabled message

ruairiphackett:
Fix for https://github.com/Shopify/shopify-themes/issues/3355

I was a little unsure of where to locate the js for the cart or whether or not to make cart into a section.


ruairiphackett:
@NathanPJF I went with the generic classes `no-cookies`/`cookies` and making `slate/cart.js` return a boolean. I removed the BEM classes and `cart.scss` as it's no longer needed.

Do you think I should keep `cart.js` under `slate/` or should it be `templates/`?

src/styles/global/helper-classes.scss
```diff
@@ -40,6 +40,22 @@
   }
 }
 
+// Only show when cookies are not supported
+.no-cookies:not(html) {
```
NathanPJF: I feel that a `cookies` class is too generic; however, we made a change in class names for our most recent theme along these lines already.  To express support for JS in a class that made more sense we switched to: `supports-js` and `supports-no-js`.  

Following the same logic:

```scss
// Only show when browser cookies are not supported
.supports-no-cookies:not(html) {
  display: none;
  
  html.supports-no-cookies & {
    display: block;
  }
}

// Only show when browser cookies are supported
.supports-cookies {
  html.supports-no-cookies & {
    display: none;
  }
}
```

src/templates/cart.liquid
```diff
@@ -102,6 +102,19 @@
...
+  {% comment %}
+  Cart Empty State
+  {% endcomment %}
+  <div class="cookies">
+    <p>{{ 'cart.general.empty' | t }}</p>
+    <p>{{ 'cart.general.continue_browsing_html' | t }}</p>
+  </div>
+
+  {% comment %}
+  Cart no cookies state
```
NathanPJF: Please put a small blurb in here that explains browser cookies are required to build a cart

src/templates/cart.liquid
```diff
@@ -102,6 +102,19 @@
   </form>
 {% else %}
   <h1>{{ 'cart.general.title' | t }}</h1>
-  <p>{{ 'cart.general.empty' | t }}</p>
-  <p>{{ 'cart.general.continue_browsing_html' | t }}</p>
+
+  {% comment %}
+  Cart Empty State
```
NathanPJF: `Cart empty state`

^ letter-casing

src/scripts/theme.js
```diff
@@ -31,4 +32,9 @@ $(document).ready(function() {
   // Wrap videos in div to force responsive layout.
   slate.rte.wrapTable();
   slate.rte.iframeReset();
+
+  // Add a message to the cart if cookies are disabled.
```
NathanPJF: This function won't directly display the message, that's taken care of by CSS.  Suggested reword:

> Apply a specific class to the html element for browser support of cookies.

src/scripts/slate/cart.js
```diff
@@ -0,0 +1,23 @@
...
+ * ------------------------------------------------------------------------------
+ * A file that contains scripts highly couple code to the Cart template.
+ *
+ * @namespace cart
+ */
+
+slate.cart = {
+  
+  /**
+   * Check if cookies are enabled in the browser
```
NathanPJF: Please put a small blurb in here that explains browser cookies are required to build a cart

ruairiphackett:
@NathanPJF I changed the class names and added the explanations you requested, thanks for the quick review 👍

Issue #96, opened 2017-02-20
Compare at price display


NathanPJF:
`product.liquid` has the following conditional for displaying the compare at price:

```
{% if current_variant.compare_at_price_max > current_variant.price %}
```
Line: https://github.com/Shopify/slate/blob/f4c6f1e648d7b7e3c9cfc037238de36b321544db/src/sections/product.liquid#L79

`.compare_at_price_max` is not a property of the `variant` object.  It should be : `current_variant.compare_at_price`

```
{% if current_variant.compare_at_price > current_variant.price %}
```

NathanPJF:
**Update**: Other issues with compare at price display:

### Compare at price on product page

The logic to display the `compare_at_price `is problematic.  It is only checking the _current variant_ `compare_at_price`.  This means that if the current variant isn't on sale, we never get the necessary markup to handle displaying compare_at_price later. 
 Line: https://github.com/Shopify/slate/blob/f4c6f1e648d7b7e3c9cfc037238de36b321544db/src/sections/product.liquid#L79

The `product.liquid` page should be using the same logic as the `featured_product` section and checking the `product` compare prices.  Then JS can handle displaying the compare at price.

### Initial compare at price display

The markup for the featured product page is showing `product.compare_at_price` at the initial render - even when that compare_at_price has nothing to do with the selected variant.

Video: http://take.ms/C8Hnv

In the video, you can see that the variant `Baby Blue / 8 / Suede` is showing a compare at price initially.  Problem: this variant is not on sale.  When I switch variants and then switch back, the proper price is shown.

Solution: the initial render should be checking the `current_variant` inside of the `product.compare_at_price` check. 

<details><summary>Example here</summary>

```
        {% if product.compare_at_price_max > product.price %}
          <span class="visually-hidden">{{ 'products.product.regular_price' | t }}</span>
          <s data-compare-price>
            {% if current_variant.compare_at_price %}{{ current_variant.compare_at_price | money }}{% endif %}
          </s>
        {% endif %}
```

</details>
&nbsp;

This will have to be considered for both the product template and the featured-product section.

### Other issues

`roduct.compare_at_price` is undocumented.

PR #111, opened 2017-03-30
Compare at price logic

NathanPJF:

Fixes https://github.com/Shopify/slate/issues/96

* Always render markup for compare at price, so JS can update it.
* On featured product page, use `current_variant` information so the initial prices displayed reflect the option selectors.

New stuff:
The changes in `product.js` have to do with how we are handling compare_at information when the variant is on sale.  We are always leaving the "compare_at" text on screen for screenreaders, even when we were hiding the sale price. Now it hides that text if the variant isn't on sale.

I also update the strikethrough price when the variant isn't on sale.  It was already being hidden with the `hide` class, but it just felt odd to leave a strikethrough price in the markup if we had no intention of using it.

/cc @Shopify/themes-fed 

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/src/scripts/sections/product.js
+++ b/src/scripts/sections/product.js
@@ -12,6 +12,7 @@ theme.Product = (function() {
     addToCart: '[data-add-to-cart]',
     addToCartText: '[data-add-to-cart-text]',
     comparePrice: '[data-compare-price]',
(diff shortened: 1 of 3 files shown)


NathanPJF:
Same approach approved here: https://github.com/Shopify/narrative/pull/114

PR #114, opened 2017-04-07
Add defaultTo function

### LINKED ISSUE #113: Currency formatting without decimals
### Problem

Using slate.Currency's formatMoney method incorrectly returns cents and decimal places if you use the `amount_no_decimals` or `amount_no_decimals_with_comma_separator` format.

### Scope

This is an issue with Slate's JavaScript functionality in the theme scaffolding - [/src/scripts/slate/currency.js](https://github.com/Shopify/slate/blob/master/src/scripts/slate/currency.js)

### Replication steps

If the `amount_no_decimals` or `amount_no_decimals_with_comma_separator` case is used `formatWithDelimiters(cents, 0);` or `formatWithDelimiters(cents, 0, '.', ',');` is called.

In the `formatWithDelimiters` method, the `precision` variable is set to 2 as a default if the variable does not exist.
```js
precision = precision || 2;
```

However, since the `precision` argument in the `formatWithDelimiters` method could be 0, the line above is seeing `precision == false == 0` it is incorrectly setting the value to 2.

### More Information

The previous version of this file [carolineschnapp/currencies/master/jquery.currencies.js](https://github.com/carolineschnapp/currencies/blob/master/jquery.currencies.js#L543-L547) used a more explicit way to handle default values.

```js
function defaultOption(opt, def) {
  return (typeof opt == 'undefined' ? def : opt);
}

precision = defaultOption(precision, 2);
```

With this explicit approach, the value of `precision` could be correctly set to `0`.

jonathanmoore:

Fixes #113 
Following the lead of PR https://github.com/Shopify/slate/pull/78, Lodash's `_.defaultTo` function was added to `slate.utils`

Now the `formatWithDelimiters` function inside `slate.Currency` can correctly apply a default value of `2` if the `precision` argument equals `0`—as well as any other place where a default variable needs to be set.

https://github.com/lodash/lodash/blob/master/defaultTo.js

```js
slate.utils.defaultTo(1, 10)
// => 1

slate.utils.defaultTo(0, 10)
// => 0

slate.utils.defaultTo(undefined, 10)
// => 10
```

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/src/scripts/slate/utils.js
+++ b/src/scripts/slate/utils.js
@@ -63,4 +63,19 @@ slate.utils = {
     }
     return result;
   }
+
+  /**
+   * _.defaultTo from lodash
+   * Checks `value` to determine whether a default value should be returned in
+   * its place. The `defaultValue` is returned if `value` is `NaN`, `null`,
+   * or `undefined`.
+   * Source: https://github.com/lodash/lodash/blob/master/defaultTo.js
+   *
+   * @param {*} value - Value to check
+   * @param {*} defaultValue - Default value
+   * @returns {*} - Returns the resolved value
+   */
+  function defaultTo(value, defaultValue) {
+    return (value == null || value !== value) ? defaultValue : value
+  }  
 };
--- a/src/scripts/slate/utils.js
+++ b/src/scripts/slate/utils.js
@@ -62,7 +62,7 @@ slate.utils = {
       }
     }
     return result;
-  }
+  },
 
   /**
    * _.defaultTo from lodash
@@ -75,7 +75,7 @@ slate.utils = {
    * @param {*} defaultValue - Default value
    * @returns {*} - Returns the resolved value
    */
-  function defaultTo(value, defaultValue) {
+  defaultTo: function(value, defaultValue) {
     return (value == null || value !== value) ? defaultValue : value
-  }  
+  }
 };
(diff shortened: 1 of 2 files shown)


src/scripts/slate/utils.js
```diff
@@ -63,4 +63,19 @@ slate.utils = {
...
+   * Checks `value` to determine whether a default value should be returned in
+   * its place. The `defaultValue` is returned if `value` is `NaN`, `null`,
+   * or `undefined`.
+   * Source: https://github.com/lodash/lodash/blob/master/defaultTo.js
+   *
+   * @param {*} value - Value to check
+   * @param {*} defaultValue - Default value
+   * @returns {*} - Returns the resolved value
+   */
+  function defaultTo(value, defaultValue) {
```
NathanPJF: This will need to be defined as a property of `utils`.

```js
defaultTo: function(value, defaultValue) {
  return (value == null || value !== value) ? defaultValue : value
}
```

src/scripts/slate/utils.js
```diff
@@ -63,4 +63,19 @@ slate.utils = {
     }
     return result;
   }
+
```
NathanPJF: need a comma after the `}` above to add another property to `slate.utils`

jonathanmoore:
Thanks for catching that! This should take care of it.

NathanPJF:
Awesome! Thanks for both reporting the issue and getting the PR in @jonathanmoore

jonathanmoore:
:raised_hands: I'm glad to help! I've been using Slate for the last few months when I converted our [premium theme District](https://themes.shopify.com/themes/district/styles/district) over to support sections. It has been tremendously helpful.

Issue #126, opened 2017-04-19
Ship with .gitignore


peterberwind:
### Problem

When creating a new theme or migrating an exsisting, there is no .gitignore so both `node_modules` and `dist` get commited to the git repo. Would be a nice to ship with this configured!

NathanPJF:
Fixed in https://github.com/Shopify/slate-cli/pull/132 

Great suggestion, @peterberwind.  This will be available in the next Slate release. Going to update the docs now.

PR #134, opened 2017-04-23
Add .gitignore for new projects

adrhumphreys:
Added git ignore, to ignore the node_modules and dist folder for new projects

This adds a gitignore file by default so that node_modules and dist don't get commited to sc by accident

[#126](https://github.com/Shopify/slate/issues/126)

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1,4 +1,4 @@
 # Custom #
 ###################
-node_modules
-dist
+node_modules/
+dist/
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1,0 +1,4 @@
+# Custom #
+###################
+node_modules
+dist


src/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Custom #
+###################
+node_modules
+dist
```
chrisberthe: Can you add a trailing slash to these to signify they're directories?

chrisberthe:
Thanks @adrhumphreys for the PR - just a small comment. @NathanPJF I've also assigned you for some 👀.

src/.gitignore
adrhumphreys: Woops, added them in there 👍

adrhumphreys:
Awesome, @chrisberthe I've add the trailing slashes

t-kelly:
Hey @adrhumphreys 👋 - have you had a chance to test this out? The `.gitignore` file you've added won't do anything because the `dist/` and `node_modules/` don't live inside the `src/` folder, they live inside the root folder. We have a `.gitignore` file for the root directory that ignores both `node_modules/` and `dist/`.

Is the existing setup not working for you? Care to elaborate as to how you ended up with `dist/` and `node_modules/` in the `src/` folder?

t-kelly:
Looks like the issue is that `.gitignore` is not being copied over when we create a new theme.

chrisberthe:
As @t-kelly mentioned, this doesn't solve the issue of ignoring the `node_modules` and `dist` folder - the `.gitignore` should be placed inside of the root theme directory, not `src`.

Closing this PR for now, but we'll definitely revisit this soon - maybe we could initialize this with a flag upon running `slate theme` and/or `slate migrate`... but don't necessarily want to clutter up options for each of those commands either. We'll 🤔.

PR #151, opened 2017-05-11
Remove Extra 'two' in theme docs

jgodson:

Just noticed an extra word while reading the docs. Removed it!


NathanPJF:
This was an homage.

PR #155, opened 2017-05-16
Update for accessible payment icons

### LINKED ISSUE #86: accessible payment icons
## Issue 
Our payment icons are unreachable by screen readers. 

### Reason

These SVGs have the attribute `aria-hidden="true"`.  [Screenshot here](http://take.ms/zecT3)

In the utility tasks, we are wrapping _all_ icons in `aria-hidden="true" focusable="false" role="presentation"`. Source: 
https://github.com/Shopify/slate-tools/blob/937bc5b76a12eaae29bbedc47ce4a6f86bc48b70/src/tasks/includes/utilities.js#L67

### Accessible SVG
For the SVG to be discoverable by screen readers, it would need to have `role="img"` and either a `title` attribute on the `svg` or a `<title>` element within the `svg`.  The `aria-hidden` would need to be removed as well.  [Screenshot](http://take.ms/FLZVN)

This would also require changing our build tasks to be able to set the `role` on an icon by icon basis.  This makes it a bit more complicated.

### Fallback text
Have "fallback text" with the class of `visually-hidden`.  [Screenshot](http://take.ms/aYUTq)  This would be simpler and we can include the HTML/Liquid in the footer section.

chrisberthe:

Fixes #86. The output for the payment icons in the footer now look like:

image

### DIFF
--- a/src/locales/de.json
+++ b/src/locales/de.json
@@ -235,7 +235,8 @@
     },
     "footer": {
       "social_platform": "{{ name }} auf {{ platform }}",
-      "copyright": "Urheberrecht"
+      "copyright": "Urheberrecht",
+      "payment_methods": "Akzeptierte Zahlungsarten"
     }
   },
   "products": {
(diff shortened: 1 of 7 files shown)


chrisberthe:
Sweet, I like the idea - I added the changes based on your feedback.

PR #161, opened 2017-05-19
Fix build tools doc

### LINKED ISSUE #82: No build tools flow
In the getting started section there’s a quote that says:

> Node 4+ is supported, but 6+ is recommended to fully benefit from Slate. If you want the template files without the build tools, get the latest zip here.

### Issue

 This download gives you the `src` files; not a distribution folder.  So you would still need the build tools installed to use it.

1. This is confusing - do I need the build tools or not?
2. Should the link be to an already compiled theme? i.e. The `dist` version?

### Suggestion

* Re-write this section as two separate thoughts.
* Clarify what the zip you're downloading is for (two options below)

---

## Getting started

> Node 4+ is supported, but 6+ is recommended to fully benefit from Slate.

The steps below walk through installing Slate globally and configuring your development theme to sync with your store.  You can `[download a zip here](link to src files)` of just the theme files, however you will require Slate's build tools to compile theme into a working theme.

-- or --

The steps below walk through installing Slate globally and configuring your development theme to sync with your store.  If you want the theme files without the build tools dependency, `[get the latest zip here](link to dist files)`.

chrisberthe:
Fixes #82.

Cleans up the build script (in case we need to create additional zips in the future). Also updates the description in the documentation for the direct zip download to the compiled `dist` version of Slate (rather than linking to `src`).

@t-kelly


scripts/build-src
```diff
@@ -1,37 +0,0 @@
...
-const SRC_DIRS = [
-  'src'
-];
-
-module.exports = (function() {
-  const output = fs.createWriteStream('upload/slate-src.zip');
-  const archive = archiver('zip');
-
-  output.on('close', () => {
-    console.log(`${archive.pointer()} total bites`);
```
chrisberthe: RIP "bites".

scripts/build
```diff
@@ -0,0 +1,33 @@
...
+
+const fs = require('fs');
+const archiver = require('archiver');
+
+module.exports = (
+  buildZip('slate-src', ['src'], ['config-sample.yml'])
+);
+
+function buildZip(name, directories = [], files = []) {
+  const output = fs.createWriteStream(`upload/${name}.zip`);
```
macdonaldr93: I got `Error: ENOENT: no such file or directory, open 'upload/slate-src.zip'` here... so I didn't have an `upload` directory. Once I added it, it was fine. So maybe check for that dir first? You can use `stat` to see if it's a directory or not.

chrisberthe: Nice! Thanks for pointing this out - will fix in the next commit.

PR #163, opened 2017-05-22
remove theme-support email, not theirs to support

NathanPJF:

Update the `theme_info` to remove Theme Support as the support email.  That group is not responsible for maintaining the code base.  

Switched it for `theme_support_url` so theme devs still know that you can separate the ideas of documentations and support.

Changed the theme version number to represent what we do for internally developed themes where the `theme_version` matches the `package.json` version. 

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/src/config/settings_schema.json
+++ b/src/config/settings_schema.json
@@ -2,10 +2,10 @@
   {
     "name": "theme_info",
     "theme_name": "Slate",
-    "theme_version": "1.0.0",
+    "theme_version": "0.10.0",
     "theme_author": "Shopify",
     "theme_documentation_url": "https://github.com/Shopify/slate",
-    "theme_support_email": "theme-support@shopify.com"
+    "theme_support_url": "https://shopify.github.io/slate/"
   },
   {
     "name": "Typography",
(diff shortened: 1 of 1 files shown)


src/config/settings_schema.json
```diff
@@ -2,10 +2,10 @@
   {
     "name": "theme_info",
     "theme_name": "Slate",
-    "theme_version": "1.0.0",
+    "theme_version": "0.10.0",
     "theme_author": "Shopify",
     "theme_documentation_url": "https://github.com/Shopify/slate",
-    "theme_support_email": "theme-support@shopify.com"
+    "theme_support_url": "https://shopify.github.io/slate/"
```
chrisberthe: Should this just be `theme_support_email: slate@shopify.com` instead? The `theme_documentation_url` is already there above.

NathanPJF: Conscious choice.  I don't want our email showing up in someone's production site 😝 

If anything, I should switch the URLs for documentation and support.  Gonna do that now.

chrisberthe: Good call 🥇

PR #164, opened 2017-05-23
use product drop for detecting default variant

NathanPJF:

There is a new product drop used for detecting whether a product only has the default variant: `product.has_only_default_variant`.  (Public docs are being updated and in review)

This PR removes the complicated and fragile check for "default product option" values.  This detection is now moved to Shopify's side.

#### Other notes

I removed `{% if option.name == 'default' %}class="label-hidden"{% endif %}` on the `<label>` elements.  Reason: If the merchant has left the product option as "Default" in the admin, then we'll display it here.  The new drop will take care of the single default variant check earlier in the markup.

### Checklist
For contributors:

For maintainers:


ConduciveMammal:
Is this now part of Shopify's default liquid?

PR #196, opened 2017-07-11
Ignore package-lock.json and npm-shrinkwrap.json

### LINKED ISSUE #195: Add `package-lock.json` and `npm-shrinkwrap.json` to `.gitignore`
### Problem

@davidcornu pointed out that `package-lock.json` [should not be included in the repo](https://github.com/Shopify/slate-tools/pull/73#issuecomment-313534803).

t-kelly:
Fixes #195


NathanPJF:
@t-kelly The `package-lock.json` is supposed to be committed to repos.  From the file's description on npmjs.org https://docs.npmjs.com/files/package-lock.json#description

>  It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.
>
> This file is **intended to be committed** into source repositories, and serves various purposes

What was pointed out in @davidcornu's comment was that this file cannot be published to NPM.  

> One key detail about `package-lock.json` is that it cannot be published, and it will be ignored if found in any place other than the toplevel package.

That's not saying it shouldn't be in a repo.  It's saying that you cannot download it from the hosted module because it's ignored on upload - which is totally fine.  The purpose of `package-lock.json` is "to ensure that the dependencies remain the same on all machines the project is installed on" (source [Beginner's guide to NPM](https://www.sitepoint.com/beginners-guide-node-package-manager/))

When you or I run `npm i`, we want it to be running from `package-lock.json` to avoid headaches experienced from relying on the dependencies expressed in `package.json` that may be too lenient.

---

The `npm-shrinkwrap.json` file is a different story since it _will be published_ to the registry, so we need to make sure we're using it right.  This Slate repo, the scaffolding of the Slate starter theme, is never being published to npmjs.org, so it's a bit of a moot point.  You can include it, but I don't forsee it being an issue.

Issue #172, opened 2017-06-06
align_logo is not a setting in header.liquid


ry167:
### Problem

[src/sections/header.liquid](https://github.com/Shopify/slate/blob/d7b02d579214ac4777eaf4f378e1e19d29ae2373/src/sections/header.liquid) contains the following code:
```
  {% if section.settings.align_logo == 'left' %}
    .site-header__logo-image {
      margin: 0;
    }
  {% endif %}
There is no matching section settings entry for this.
```

Should this be added as a setting or should this code be removed?

NathanPJF:
@ry167 Good catch!  This code seems to be left over from Debut.  There shouldn't be any reference to a `section.settings.align_logo` in the Slate theme.

NathanPJF:
Related:  `logo_max_width` should be using the new `range` input instead of `text`

PR #198, opened 2017-07-13
Update settings for header section

t-kelly:
Fixes https://github.com/Shopify/slate/issues/172

For maintainers:

### DIFF
--- a/src/sections/header.liquid
+++ b/src/sections/header.liquid
@@ -4,12 +4,6 @@
       max-width: {{ section.settings.logo_max_width }}px;
     }
   {% endif %}
-
-  {% if section.settings.align_logo == 'left' %}
-    .site-header__logo-image {
-      margin: 0;
-    }
-  {% endif %}
 </style>
 <div data-section-id="{{ section.id }}" data-section-type="header-section">
   <header role="banner">
@@ -121,10 +115,14 @@
         "label": "Logo image"
       },
       {
-        "type": "text",
+        "type": "range",
         "id": "logo_max_width",
+        "min": 50,
+        "max": 450,
+        "step": 8,
+        "unit": "px",
         "label": "Custom logo width (in pixels)",
-        "default": "250"
+        "default": 250
       },
       {
         "type": "link_list",
--- a/src/sections/header.liquid
+++ b/src/sections/header.liquid
@@ -119,9 +119,9 @@
         "id": "logo_max_width",
         "min": 50,
         "max": 450,
-        "step": 8,
+        "step": 10,
         "unit": "px",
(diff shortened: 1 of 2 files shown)


chrisberthe:
Looks right - do you have a demo shop we could test?

src/sections/header.liquid
```diff
@@ -121,10 +115,14 @@
...
       },
       {
-        "type": "text",
+        "type": "range",
         "id": "logo_max_width",
+        "min": 50,
+        "max": 450,
+        "step": 8,
+        "unit": "px",
         "label": "Custom logo width (in pixels)",
```
chrisberthe: Units are there, so no need for `(in pixels)`.

src/sections/header.liquid
```diff
@@ -121,10 +115,14 @@
         "label": "Logo image"
       },
       {
-        "type": "text",
+        "type": "range",
         "id": "logo_max_width",
+        "min": 50,
+        "max": 450,
+        "step": 8,
```
chrisberthe: Is 8 a generic value you chose or one of our standards now?

chrisberthe: Is 8 a generic value you chose or one of our standards now?

t-kelly: Is there another theme that is using the range selector for the logo yet?

chrisberthe: No, it seems like there aren't any yet for headers. I understood how you got 8 but was just wondering if we had established a pattern for this. Bump it to 10? 😛

t-kelly:
> You'll need to add width: 100% on the image for it to scale properly.

Do we want to stretch the logo? Isn't that what would happen if we added `width: 100%;`?

src/sections/header.liquid
t-kelly: Sure 👍

chrisberthe:
Not sure, realistically we wouldn't want to stretch the image. I just found it weird that with a smaller image it caps out at its actual width even if I'm selecting "400px". I'd be curious what others think.

t-kelly:
> I'd be curious what others think.
@NathanPJF @MaxMusing @huguestennier ^^^

huguestennier:
I feel like if we cap it then there's no point since it could just go smaller.

Hail mary, but isn't that against performance to load a big image for logo and make it smaller with the range slider? Shouldn't we encourage merchants to upload to the right size?

huguestennier:
I feel like we should let it resize bigger even if it gets blurry because I wouldn't want the range slider to not work at some value since the image as a max-width.

The client can always upload a better image.

Issue #214, opened 2017-08-09
og:image no longer works for Facebook sharing


itsjustmath:
### Problem

A client was trying to share their newly launched store via Facebook, but the it wouldn't show an image preview.

![2017-08-09 at 1 54 pm](https://user-images.githubusercontent.com/635044/29143530-63f77c4e-7d0a-11e7-8f88-ae3a0d720d51.png)

Upon further inspection, it looks like this was an issue in `social-meta-tags.liquid` with the [`og:image`](http://ogp.me/#structured) meta property.

### Scope

Slate's theme scaffolding => `src/snippets/social-meta-tags.liquid`

### Replication steps

The issue has been resolved for our client, but you can pass a link to Facebook's [sharing debugger](https://developers.facebook.com/tools/debug/sharing/), for various template types (I did: homepage, a collections page, and a product page). 

Here's an example of the `og:image:url` working properly: https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fwww.bucketfeet.com%2Fcollections%2Fwomen%2Fproducts%2Fthe-duo-unisex-1

### More Information

Updating it to `og:image:url` resolved the issue for us.

NathanPJF:
Thanks for the PR, @itsjustmath 

I think this is a bug on Facebook's end.  In the scrapper link you provided, you can see that the `og:type` is being labeled as `website` - despite the fact the the `type` is clearly marked as `product`.  Screenshot here: http://take.ms/5N0gy.  This is why the warnings are also saying that you have extraneous properties like price, http://take.ms/gw8tH

I was getting the same behaviour on my end when I was testing.  It wasn't until I ran my own link through [Open Graph Object Debugger](https://developers.facebook.com/tools/debug/og/object/) that I finally saw a change.  Screenshot of debugger: http://take.ms/4xE2Y .

Once I did that, the sharing URL finally showed the `type` as `product` and `og:image` started working again.  Screenshot: http://take.ms/1BFcl 

Buuuut, I don't even think running it through the debugger helped.  Because I tried to reproduce on another test store.  It said it couldn't find `og:image:url`, so I scraped again and suddenly it was happy with `og:image` even though the `type` was still `website`. Screenshot: http://take.ms/mda3Y.  I think it's just a timing thing for stuff I scrape for the first time; the debugger even gave me a warning that images are fetched asynchronously, so I won't see it right away for stuff that hasn't been scraped before.

[...]

PR #215, opened 2017-08-09
Update og:image meta property to og:image:url

itsjustmath:

- [x] Update `og:image` meta property to `og:image:url` (bug noted in #214)
  - [Working example](https://developers.facebook.com/tools/debug/sharing/?q=https%3A%2F%2Fwww.bucketfeet.com%2Fcollections%2Fwomen%2Fproducts%2Fthe-duo-unisex-1)

### Checklist
For contributors:

### DIFF
--- a/src/snippets/social-meta-tags.liquid
+++ b/src/snippets/social-meta-tags.liquid
@@ -11,7 +11,7 @@
   {%- assign og_type = 'product' -%}
   {%- capture og_image_tags -%}
     {%- for image in product.images limit: 3 -%}
-      <meta property="og:image" content="http:{{ image.src | product_img_url: '1024x1024' }}">
+      <meta property="og:image:url" content="http:{{ image.src | product_img_url: '1024x1024' }}">
     {%- endfor -%}
   {%- endcapture -%}
   {%- capture og_image_secure_url_tags -%}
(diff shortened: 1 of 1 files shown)


NathanPJF:
Thanks for the PR, @itsjustmath 

I think this is a bug on Facebook's end.  In the scrapper link you provided, you can see that the `og:type` is being labeled as `website` - despite the fact the the `type` is clearly marked as `product`.  Screenshot here: http://take.ms/5N0gy.  This is why the warnings are also saying that you have extraneous properties like price, http://take.ms/gw8tH

I was getting the same behaviour on my end when I was testing.  It wasn't until I ran my own link through [Open Graph Object Debugger](https://developers.facebook.com/tools/debug/og/object/) that I finally saw a change.  Screenshot of debugger: http://take.ms/4xE2Y .

Once I did that, the sharing URL finally showed the `type` as `product` and `og:image` started working again.  Screenshot: http://take.ms/1BFcl 

Buuuut, I don't even think running it through the debugger helped.  Because I tried to reproduce on another test store.  It said it couldn't find `og:image:url`, so I scraped again and suddenly it was happy with `og:image` even though the `type` was still `website`. Screenshot: http://take.ms/mda3Y.  I think it's just a timing thing for stuff I scrape for the first time; the debugger even gave me a warning that images are fetched asynchronously, so I won't see it right away for stuff that hasn't been scraped before.

Also, it looks like the site you linked is using `og:image` and not `og:image:url`, screenshot: http://take.ms/gmpXs . Any reason you switched it back to `og:image`?

itsjustmath:
@NathanPJF thanks looking into this and getting back to me. I'll close this and #214 for now!

As for the discrepancies with product pages showing  `og:image` - the client may not have deployed that fix to their store yet.

I'll try reverting the appended `:url`, and I'll report back if I run into any issues.

Issue #99, opened 2017-02-23
product view and JS require an image


NathanPJF:
## Issue

The markup and JS assumes your product images will always have an image.  
* When product template doesn't have a featured image, it renders the placeholder tag.
* The JS throws an error when it goes looking for an image selector and doesn't find one.

I realize it's not common for a product not to have an image, but it's an assumption baked into the theme that should be addressed.

### Reproduction

Make a product that has no images associated with it.  Example screenshot: http://take.ms/zIVff

<details><summary>Screenshot of feature product section</summary>

![http://take.ms/P5jDA](http://take.ms/P5jDA)

</details>

### Problem code
Three areas:
* `featured-product.liquid`
* `product.liquid`
* `product.js`

In `featured-product.liquid`, it assumes that if you don't have a featured image then you must be onboarding:
```
  {% if product.featured_image.src != blank %}
    {{ product.featured_image.src | img_url: '480x480' | img_tag: product.featured_image.src.alt, 'product-image' }}
  {% else %}
    {{ 'product-1' | placeholder_svg_tag: 'placeholder-svg placeholder-svg--small' }}
  {% endif %}
```

In `product.liquid` there's no check if a featured image exists.

In `product.js` there is no check whether `$(selectors.productFeaturedImage, this.$container)` yields any results first.

```js
this.settings.imageSize = slate.Image.imageSize($(selectors.productFeaturedImage, this.$container).attr('src'))
```

NathanPJF:
Possibly related: https://github.com/Shopify/narrative/pull/252/

PR #239, opened 2017-08-18
Fix behaviour and errors when no product image

t-kelly:

Fixes https://github.com/Shopify/slate/issues/99

- Fixes `feature-product.liquid` section by removing placeholder image if no image is set. Also matches `product.liquid` section behaviour by setting the featured image as either the current variant's featured image or the product featured image
- Fixes `product.liquid` by checking if featured image is not black
- Fixes `product.js` by only performing image functions if an image is present on the page. There was some code that was rearranged to make this check easy.

For maintainers:

### DIFF
--- a/packages/slate-theme/src/scripts/sections/product.js
+++ b/packages/slate-theme/src/scripts/sections/product.js
@@ -37,7 +37,6 @@ theme.Product = (function() {
     }
 
     var sectionId = this.$container.attr('data-section-id');
-    var $featuredImage = $(selectors.productFeaturedImage, this.$container);
     var options = {
       $container: this.$container,
       enableHistoryState: this.$container.data('enable-history-state') || false,
@@ -49,13 +48,14 @@ theme.Product = (function() {
     this.settings = {};
     this.namespace = '.product';
     this.variants = new slate.Variants(options);
+    this.$featuredImage = $(selectors.productFeaturedImage, this.$container);
     this.productSingleObject = JSON.parse($(selectors.productJson, this.$container).html());
 
     this.$container.on('variantChange' + this.namespace, this.updateAddToCartState.bind(this));
     this.$container.on('variantPriceChange' + this.namespace, this.updateProductPrices.bind(this));
 
-    if ($featuredImage.length > 0) {
-      this.settings.imageSize = slate.Image.imageSize($(selectors.productFeaturedImage, this.$container).attr('src'));
+    if (this.$featuredImage.length > 0) {
+      this.settings.imageSize = slate.Image.imageSize(this.$featured_image.attr('src'));
       slate.Image.preload(this.productSingleObject.images, this.settings.imageSize);
 
       this.$container.on('variantImageChange' + this.namespace, this.updateProductImage.bind(this));
@@ -122,9 +122,8 @@ theme.Product = (function() {
     updateProductImage: function(evt) {
       var variant = evt.variant;
       var sizedImgUrl = slate.Image.getSizedImageUrl(variant.featured_image.src, this.settings.imageSize);
-      var $image = $(selectors.productFeaturedImage, this.$container);
 
-      $(selectors.productFeaturedImage, this.$container).attr('src', sizedImgUrl);
+      this.$featured_image.attr('src', sizedImgUrl);
     },
 
     /**
--- a/packages/slate-theme/src/scripts/sections/product.js
+++ b/packages/slate-theme/src/scripts/sections/product.js
@@ -29,10 +29,6 @@ theme.Product = (function() {
    */
   function Product(container) {
     this.$container = $(container);
-    var sectionId = this.$container.attr('data-section-id');
-
-    this.settings = {};
-    this.namespace = '.product';
 
     // Stop parsing if we don't have the product json script tag when loading
     // section in the Theme Editor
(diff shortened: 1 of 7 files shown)


t-kelly:
Added some additional checks:
- Check if product image exists in the featured-collection.liquid
- Check if item image exists in the cart.liquid
- Check if featured image exists in collection.liquid
- Check if featured image exists in search.liquid

packages/slate-theme/src/scripts/sections/product.js
```diff
@@ -29,47 +29,42 @@ theme.Product = (function() {
...
     this.productSingleObject = JSON.parse($(selectors.productJson, this.$container).html());
-    this.settings.imageSize = slate.Image.imageSize($(selectors.productFeaturedImage, this.$container).attr('src'));
 
-    slate.Image.preload(this.productSingleObject.images, this.settings.imageSize);
+    this.$container.on('variantChange' + this.namespace, this.updateAddToCartState.bind(this));
+    this.$container.on('variantPriceChange' + this.namespace, this.updateProductPrices.bind(this));
 
-    this.initVariants();
+    if ($featuredImage.length > 0) {
+      this.settings.imageSize = slate.Image.imageSize($(selectors.productFeaturedImage, this.$container).attr('src'));
```
chrisberthe: The featured product is already cache above:

```js
this.settings.imageSize = slate.Image.imageSize($featuredImage.attr('src'));
```

packages/slate-theme/src/scripts/sections/product.js
```diff
@@ -127,6 +122,7 @@ theme.Product = (function() {
     updateProductImage: function(evt) {
       var variant = evt.variant;
       var sizedImgUrl = slate.Image.getSizedImageUrl(variant.featured_image.src, this.settings.imageSize);
+      var $image = $(selectors.productFeaturedImage, this.$container);
 
       $(selectors.productFeaturedImage, this.$container).attr('src', sizedImgUrl);
```
chrisberthe: Cached on line 125:

```js
$image.attr('src', sizedImgUrl);
```

Either that or omit caching it here altogether.

packages/slate-theme/src/scripts/sections/product.js
```diff
@@ -29,47 +29,42 @@ theme.Product = (function() {
...
-    this.namespace = '.product';
 
     // Stop parsing if we don't have the product json script tag when loading
     // section in the Theme Editor
     if (!$(selectors.productJson, this.$container).html()) {
       return;
     }
 
+    var sectionId = this.$container.attr('data-section-id');
+    var $featuredImage = $(selectors.productFeaturedImage, this.$container);
```
chrisberthe: Maybe we can assign this to `this.$featuredImage` considering we're accessing this element both in the constructor (if it exists) and the `updateProductImage` method below.

packages/slate-theme/src/scripts/sections/product.js
chrisberthe: If we go with `this.$featuredImage` above, replace with that.

packages/slate-theme/src/scripts/sections/product.js
```diff
@@ -127,6 +122,7 @@ theme.Product = (function() {
     updateProductImage: function(evt) {
       var variant = evt.variant;
       var sizedImgUrl = slate.Image.getSizedImageUrl(variant.featured_image.src, this.settings.imageSize);
+      var $image = $(selectors.productFeaturedImage, this.$container);
```
chrisberthe: If we go with the instance variable suggestion above, we can omit this altogether.

packages/slate-theme/src/templates/cart.liquid
```diff
@@ -24,10 +24,15 @@
...
           <tr class="responsive-table-row">
+
             <td data-label="{{ 'customer.order.product' | t }}">
-              <a href="{{ item.url | within: collections.all }}">
-                <img src="{{ item | img_url: '240x240' }}" alt="{{ item.title | escape }}">
-              </a>
+
+              {% if item.image != blank %}
+                <a href="{{ item.url | within: collections.all }}">
+                  <img src="{{ item | img_url: '240x240' }}" alt="{{ item.title | escape }}">
```
chrisberthe: There's minor inconsistencies in the way we use image tags in Slate. Above, we're using:

```
{{ product.featured_image.src | img_url: '480x480' | img_tag: product.title }}
```

While other instances we're using the HTML `` tag - most likely due to the fact that with the former we're not able to pass a data attribute as a parameter. We could update this to keep it more consistent when not using a data attribute.

packages/slate-theme/src/templates/collection.liquid
```diff
@@ -27,9 +27,11 @@
   <div>
     {% for product in collection.products %}
       <div>
-        <a href="{{ product.url | within: collection }}">
-          <img src="{{ product.featured_image.src | img_url: '480x480' }}" alt="{{ product.featured_image.alt | escape }}">
-        </a>
+        {% if product.featured_image != blank %}
+          <a href="{{ product.url | within: collection }}">
+            <img src="{{ product.featured_image.src | img_url: '480x480' }}" alt="{{ product.featured_image.alt | escape }}">
```
chrisberthe: Same here.

Issue #227, opened 2017-08-11
slate watch catches DS_Store with every file change


t-kelly:
<a href="https://github.com/l-fischer"><img src="https://avatars0.githubusercontent.com/u/1226404?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [l-fischer](https://github.com/l-fischer)**
_Thursday Aug 10, 2017 at 15:03 GMT_
_Originally opened as https://github.com/Shopify/slate-cli/issues/134_

----

Every local change while slate is watching, triggers the theme upload of the file successfully, then detects change in .DS_Store, tries to run theme upload on that file and then gets hung there, staying at 0%

![ds_store-bug](https://user-images.githubusercontent.com/1226404/29176743-bcdd725e-7dba-11e7-97ab-d75c0861f25a.jpg)

Tried adding DS_Store to ignore files and has no effect
```yml
ignore_files:
- "*.DS_Store"
```

alistairholt:
This has become a problem for me again recently. I'm running 0.12.1 from https://www.npmjs.com/package/@shopify/slate. Example output after running `slate watch` and making a change:

```
$ slate watch
[10:46:04] Using gulpfile ~/xxxx/@shopify/slate-tools/lib/gulpfile.js
[10:46:04] Starting 'watch'...
[10:46:04] Starting 'validate:id'...
[10:46:04] Finished 'watch' after 7.11 ms
[10:46:04] Finished 'validate:id' after 9.24 ms
[10:46:04] Starting 'build:config'...
[10:46:04] running task - build:config
[10:46:04] config.yml 1.06 kB
[10:46:04] Finished 'build:config' after 22 ms
[10:46:04] Starting 'watch:assets'...
[10:46:04] Finished 'watch:assets' after 7.05 ms
[10:46:04] Starting 'watch:config'...
[10:46:04] Finished 'watch:config' after 189 μs
[10:46:04] Starting 'watch:svg'...
[10:46:04] Finished 'watch:svg' after 369 μs
[10:46:04] Starting 'watch:css'...
[10:46:04] Finished 'watch:css' after 1.34 ms
[10:46:04] Starting 'watch:js'...
[10:46:04] Finished 'watch:js' after 437 μs
[10:46:04] Starting 'watch:vendor-js'...
[10:46:04] Finished 'watch:vendor-js' after 346 μs
[10:46:04] Starting 'watch:src'...
[10:46:04] Finished 'watch:src' after 24 μs
[...]

PR #240, opened 2017-08-18
Ignore watching dot files in dist/ directory

t-kelly:

Fixes https://github.com/Shopify/slate/issues/227

Ignores watching `.DS_store` and any other dot files in the `dist/` directory.

For maintainers:

### DIFF
--- a/packages/slate-tools/src/tasks/build-assets.js
+++ b/packages/slate-tools/src/tasks/build-assets.js
@@ -82,10 +82,12 @@ gulp.task('build:assets', () => {
 gulp.task('watch:assets', () => {
   const eventCache = utils.createEventCache();
 
-  chokidar.watch(assetsPaths, {ignoreInitial: true})
-    .on('all', (event, path) => {
-      messages.logFileEvent(event, path);
-      eventCache.addEvent(event, path);
-      utils.processCache(eventCache, processAssets, removeAssets);
-    });
+  chokidar.watch(assetsPaths, {
+    ignored: /(^|[/\\])\../,
+    ignoreInitial: true,
+  }).on('all', (event, path) => {
+    messages.logFileEvent(event, path);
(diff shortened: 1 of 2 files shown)


NathanPJF:
I've 🎩'd these changes.  Changes in dot-files 

Something to note: dot-files are being ignore in the `dist/` directory, however, their presence in the `src/assets` directory will still trigger the `build:asset` task.  This only seems to effect the `assets/` directory, other directories will ignore dot-files.

In the screenshot below, I experiment with adding `*.txt` files and `.DS_Store` files to the `src/assets` directory.  Note: The `.DS_Store` files are not being generated by my macOS as I have 10.12.6 installed; I'm am generating them manually with `touch .DS_Store`

You can see that in case 1 and 3, a new asset triggers a build and then deploy.  In case 2 and the last one, a `.DS_Store` file will trigger `build:assets`.



## Note about hanging spinner

A note about the hanging spinner mentioned in the original issue:

```
[development]: 0 / 0 [|]  0%
```

This was previously raised as a bug in Themekit 0.6.12.  You can find the original issue here: https://github.com/Shopify/themekit/issues/436

It appears to be addressed in an upcoming release of ThemeKit.  Until then, this is a decent patch.

NathanPJF:
Bumping to v0.7.0 shouldn't be taken lightly though.  It introduces `theme.lock` for remote file checking.  Super useful, but we'll want to make sure Slate handles and displays errors correctly for this new tool.

t-kelly:
> Something to note: dot-files are being ignore in the dist/ directory, however, their presence in the src/assets directory will still trigger the build:asset task. This only seems to effect the assets/ directory, other directories will ignore dot-files.

This makes sense since the build/watch process for `src/assets/` is to copy any file to `dist/assets`. Should we ignore all dot files in the `src/assets/` folder? I think yes.

PR #246, opened 2017-08-23
Update main navigation liquid to support 3 level

### LINKED ISSUE #237: Update Nested Navigation snippet
### Problem

Currently, Slate default "theme" is using the old `liquid` markup. We would need to update it to support 3 level and also add `js` to support basic keyboard navigation out of the box.

huguestennier:

Fixes #237.

Update nested nav liquid. As discussed IRL with @t-kelly, we're also stripping down the classes and the arrow down icon to make it as barebone as possible.

**This need to be merged only when the new liquid is rolled out 100%**

### DIFF
--- a/packages/slate-theme/src/sections/header.liquid
+++ b/packages/slate-theme/src/sections/header.liquid
@@ -78,31 +78,41 @@
   </header>
 
   <nav role="navigation">
-    <ul class="site-nav">
+    <ul>
       {% for link in linklists[section.settings.main_linklist].links %}
-        {%- assign child_list_handle = link.title | handleize -%}
-        {% if menus[child_list_handle].links != blank %}
-          <li class="site-nav--has-submenu{% if link.active %} site-nav--active{% endif %}">
-            <a href="{{ link.url }}" class="site-nav__link">
-              {{ link.title }}
-              {% include 'icon-arrow-down' %}
-            </a>
-            <ul class="site-nav__submenu">
-              {% for childlink in menus[child_list_handle].links %}
-                <li {% if childlink.active %}class="site-nav--active"{% endif %}>
-                  <a href="{{ childlink.url }}" class="site-nav__link">{{ childlink.title | escape }}</a>
-                </li>
+        {% if link.links != blank %}
+          <li>
+            <a href="{{ link.url }}">{{ link.title | escape }}</a>
+            <ul>
+              {% for childlink in link.links %}
+                {% if childlink.links != blank %}
+                  <li>
+                    <a href="{{ childlink.url }}">{{ childlink.title | escape }}</a>
+                    <ul>
+                      {% for grandchildlink in childlink.links %}
+                        <li>
+                          <a href="{{ grandchildlink.url }}">{{ grandchildlink.title | escape }}</a>
+                        </li>
+                      {% endfor %}
+                    </ul>
+                  </li>
+                {% else %}
+                  <li>
+                    <a href="{{ childlink.url }}">{{ childlink.title | escape }}</a>
+                  </li>
+                {% endif %}
               {% endfor %}
             </ul>
           </li>
         {% else %}
-          <li {% if link.active %}class="site-nav--active"{% endif %}>
-            <a href="{{ link.url }}" class="site-nav__link">{{ link.title }}</a>
+          <li>
+            <a href="{{ link.url }}">{{ link.title | escape }}</a>
           </li>
         {% endif %}
       {% endfor %}
     </ul>
   </nav>
+
 </div>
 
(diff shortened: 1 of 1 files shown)


packages/slate-theme/src/sections/header.liquid
```diff
@@ -78,31 +78,41 @@
...
-          <li {% if link.active %}class="site-nav--active"{% endif %}>
-            <a href="{{ link.url }}" class="site-nav__link">{{ link.title }}</a>
+          <li>
+            <a href="{{ link.url }}">{{ link.title | escape }}</a>
           </li>
         {% endif %}
       {% endfor %}
     </ul>
   </nav>
+
```
chrisberthe: 🔥 this line.

packages/slate-theme/src/sections/header.liquid
```diff
@@ -78,31 +78,41 @@
...
-              {{ link.title }}
-              {% include 'icon-arrow-down' %}
-            </a>
-            <ul class="site-nav__submenu">
-              {% for childlink in menus[child_list_handle].links %}
-                <li {% if childlink.active %}class="site-nav--active"{% endif %}>
-                  <a href="{{ childlink.url }}" class="site-nav__link">{{ childlink.title | escape }}</a>
-                </li>
+        {% if link.links != blank %}
+          <li>
```
chrisberthe: Instead of having the `if` condition right after the `for` loop, we can extract the `<li>` so we don't need to rewrite it in the `{% else %}` block:

```liquid
{% for link in linklists[section.settings.main_linklist].links %}
  <li>
    {% if link.links != blank %}
      ...
    {% else %}
      <a href="{{ childlink.url }}">{{ childlink.title | escape }}</a>
    {% endif %}
  </li>
{% endfor %}
```

packages/slate-theme/src/sections/header.liquid
```diff
@@ -78,31 +78,41 @@
...
+            <a href="{{ link.url }}">{{ link.title | escape }}</a>
+            <ul>
+              {% for childlink in link.links %}
+                {% if childlink.links != blank %}
+                  <li>
```
chrisberthe: Same thing applies for this one as well.

packages/slate-theme/src/sections/header.liquid
huguestennier: The `<li>` is inside the `if` because you probably want to add a special `class` or `data-attribute` to it since it contains a dropdown. But we can put it inside in that case. What do you think?

chrisberthe: Yeah good point - I was on the fence about this.

I figured because it's super boilerplate, and we're not leveraging any additional classes it stood out more. We're doing something similar on Narrative:

```liquid
<li class="nav__sublink{% if childlink.links != blank %} nav__has-sublinks{% endif %}">
```

Syntactically it ends up being personal preference once you implement a third level - I'll leave this up to you and @t-kelly.

t-kelly: Otherwise, this is valid Liquid syntax:
```html
{% for product in collections.frontpage.products %}
    {{ forloop.index }}
{% else %}
    // no products in your frontpage collection
{% endfor %}
```

chrisberthe: Good point - I always just assumed it was for bleeding edges cases for backwards compatibility with the old implementation. Wonder why we're showing `{% if link.links != blank %} on our documentation then.

NathanPJF: I'm a fan of the for loop @t-kelly is suggesting

huguestennier: I'm not sure what @t-kelly is suggesting is suppose to replace 🤔

huguestennier: After discussion, we'll keep it inside.

t-kelly:
Can we merge this now that nav has shipped?

chrisberthe:
Not yet - only when nested navigation has been rolled out completely.

NathanPJF:
@huguestennier @chrisberthe @t-kelly I have a proposal: This PR could be amended with the existing/old Liquid way included as a snippet.  The snippet would be commented out and contain text explaining that it's there for legacy reasons if you don't have the latest UI.  

And then in a future PR we can remove the commented out snippet.

huguestennier:
Just question, but are we doing a new Slate release soon? Is there any urgency to merge this PR?

t-kelly:
No urgency, just housekeeping and keeping things in the repo tidy. I'd like to do something with it. Merge it or close it until we plan on merging it.

PR #276, opened 2017-09-07
Migrate to CircleCI 2

t-kelly:

Update Slate to use CircleCI 2. Should decrease the time it takes to run.

cc @chrisberthe


t-kelly:
After testing this, CI run times have gone from ~140 seconds to ~40 seconds, a 350% decrease!

PR #280, opened 2017-09-11
Add dutch locale

driespieters:

Add dutch translation. (Double checked for spelling errors & correct context.)

### Checklist
:white_check_mark: I have updated the docs


NathanPJF:
Right on, @driespieters - thanks for doing this!  I did a double check as well as Google translate would let me 😅

PR #294, opened 2017-09-24
Rte helpers fallback class

### LINKED ISSUE #293: Rte helpers fallback class
### Problem

When using the helpers `slate.rte.wrapTable` and `slate.rte.wrapIframe`, if you do not provide an option for `tableWrapperClass` the script will apply a class of `undefined`.

### Replication steps

In `theme.js`, call `slate.rte.wrapTable` with no class option provided.

```js
  slate.rte.wrapTable({
    $tables: $(tableSelectors)
  });
```

NathanPJF:

Fixes #293

To prevent this silent failure/unintended behaviour, I've added a `typeof` check so only strings may be applied. 

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/packages/slate-theme/src/scripts/slate/rte.js
+++ b/packages/slate-theme/src/scripts/slate/rte.js
@@ -15,7 +15,7 @@ slate.rte = {
    * @param {string} options.tableWrapperClass - table wrapper class name
    */
   wrapTable: function(options) {
-    var tableWrapperClass = typeof options.tableWrapperClass === "undefined" ? '' : options.tableWrapperClass;
+    var tableWrapperClass = typeof options.tableWrapperClass !== 'string' ? '' : options.tableWrapperClass;
 
     options.$tables.wrap('<div class="' + tableWrapperClass + '"></div>');
   },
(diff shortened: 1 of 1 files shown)


packages/slate-theme/src/scripts/slate/rte.js
```diff
@@ -15,7 +15,9 @@ slate.rte = {
    * @param {string} options.tableWrapperClass - table wrapper class name
    */
   wrapTable: function(options) {
-    options.$tables.wrap('<div class="' + options.tableWrapperClass + '"></div>');
+    var tableWrapperClass = typeof options.tableWrapperClass !== 'string' ? '' : options.tableWrapperClass;
```
t-kelly: If some passes `1234` then would result in a empty string, not `'1234'`. Is this okay? I'm on the fence.

NathanPJF: That's a fair concern.  I can change it back to checking `typeof undefined` and then outputting the value always as a string.

Regarding the integer class names: Having class names that _starts_ with a number requires some work arounds to function in a stylesheet - [quick codepen here](https://codepen.io/NathanPJF/pen/rGjNxg).  But those workarounds exist, so I shouldn't discount it.

PR #309, opened 2017-10-04
Ensure that slate documentation is indexed by search engines

adrhumphreys:
### Ensure Slate documentation is indexed
Currently all the slate documentation has the meta tag "noindex" applied to it. This prevents search engines from index and showing the page in search results.

We can see that Google specifies [here](https://developers.google.com/search/reference/robots_meta_tag) what noindex means ("Do not show this page in search results and do not show a "Cached" link in search results.")

Since the default behavior is to index a page if it doesn't specify the meta robots tag we can safely just remove it

### Checklist
For contributors:


NathanPJF:
> My guess is that this was forgotten about when we switched the repo from private to public?

My guess as well 😅 

Thanks @adrhumphreys !

PR #323, opened 2017-10-16
Beautify settings files on migration

### LINKED ISSUE #224: Enhance the migrate command to beautify the settings_schema.json and settings_data.json files
<a href="https://github.com/michaelrshannon"></img></a> **Issue by [michaelrshannon](https://github.com/michaelrshannon)**
_Thursday Apr 06, 2017 at 16:15 GMT_
_Originally opened as https://github.com/Shopify/slate-cli/issues/117_

----

### Problem

When downloading an existing theme from a store, the contents of `/config` will be compressed, reducing their readability. It would be great if as part of the `migrate` command, these files could be uncompressed, to make them easy to work with.

### Scope

Most visible as part of `slate migrate`, but may involve modification to [slate-tools](https://github.com/Shopify/slate-tools).

### Replication steps

1. Download an existing theme from Shopify
2. Open `config/settings_data.json`

### More Information

Current workaround during dev is to open the `settings_*.json` files in the admin-based web editor for the theme you're working on, where they'll be shown uncompressed, and copy/paste into your local versions of those files.

maximevaillancourt:
Beautify the settings files (`settings_*.json`) upon migration (fix #224).

### Expected results
The settings files are pretty printed (indented with 2 spaces).

### Checklist
For contributors:
- [X] I have [updated the docs](https://github.com/Shopify/slate/blob/master/CONTRIBUTING.md#documentation) to reflect these changes, if applicable.

For maintainers:
- [X] I have :tophat:'d these changes.

### DIFF
--- a/packages/slate-cli/src/commands/migrate.js
+++ b/packages/slate-cli/src/commands/migrate.js
@@ -4,7 +4,7 @@ import {join} from 'path';
 import {prompt} from 'inquirer';
 import {green, red, yellow} from 'chalk';
 import figures from 'figures';
-import {downloadFromUrl, startProcess, writePackageJsonSync, move, isShopifyTheme, isShopifyThemeWhitelistedDir} from '../utils';
+import {downloadFromUrl, startProcess, writePackageJsonSync, move, beautifyJson, isShopifyTheme, isShopifyThemeSettingsFile, isShopifyThemeWhitelistedDir} from '../utils';
 
 export default function(program) {
   program
@@ -35,6 +35,7 @@ export default function(program) {
       const configYml = join(workingDirectory, 'config.yml');
       const pkgJson = join(workingDirectory, 'package.json');
       const srcDir = join(workingDirectory, 'src');
+      const srcConfigDir = join(workingDirectory, 'src/config');
       const iconsDir = join(srcDir, 'icons');
       const stylesDir = join(srcDir, 'styles');
       const scriptsDir = join(srcDir, 'scripts');
@@ -73,14 +74,29 @@ export default function(program) {
 
       const files = readdirSync(workingDirectory);
       const whitelistFiles = files.filter(isShopifyThemeWhitelistedDir);
-      const promises = whitelistFiles.map(movePromiseFactory);
+      const movePromises = whitelistFiles.map(movePromiseFactory);
+
+      function beautifyJsonPromiseFactory(file) {
+        console.log(`  Beautifying ${file}...`);
+        return beautifyJson(join(srcConfigDir, file));
+      }
+
+      const configDirFiles = readdirSync(srcConfigDir);
+      const themeSettingsFiles = configDirFiles.filter(isShopifyThemeSettingsFile);
+      const beautifyPromises = themeSettingsFiles.map(beautifyJsonPromiseFactory);
 
       try {
-        await Promise.all(promises);
+        await Promise.all(movePromises);
 
         console.log('');
         console.log(`  ${green(figures.tick)} Migration to src/ completed`);
         console.log('');
+
+        await Promise.all(beautifyPromises);
+
+        console.log(`  ${green(figures.tick)} Beautification of settings files completed`);
+        console.log('');
+
         console.log('  Installing Slate dependencies...');
         console.log('');
 
--- a/packages/slate-cli/src/commands/migrate.js
+++ b/packages/slate-cli/src/commands/migrate.js
@@ -4,7 +4,7 @@ import {join} from 'path';
 import {prompt} from 'inquirer';
 import {green, red, yellow} from 'chalk';
 import figures from 'figures';
-import {downloadFromUrl, startProcess, writePackageJsonSync, move, beautifyJson, isShopifyTheme, isShopifyThemeSettingsFile, isShopifyThemeWhitelistedDir} from '../utils';
+import * as utils from '../utils';
 
 export default function(program) {
   program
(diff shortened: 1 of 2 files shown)


packages/slate-cli/src/commands/migrate.js
```diff
@@ -4,7 +4,7 @@ import {join} from 'path';
 import {prompt} from 'inquirer';
 import {green, red, yellow} from 'chalk';
 import figures from 'figures';
-import {downloadFromUrl, startProcess, writePackageJsonSync, move, isShopifyTheme, isShopifyThemeWhitelistedDir} from '../utils';
+import {downloadFromUrl, startProcess, writePackageJsonSync, move, beautifyJson, isShopifyTheme, isShopifyThemeSettingsFile, isShopifyThemeWhitelistedDir} from '../utils';
```
t-kelly: I feel like this declaration is overkill -- what's your opinion on importing the whole `utils` object and call the methods from that?

packages/slate-cli/src/utils.js
```diff
@@ -178,3 +197,13 @@ export function isShopifyThemeWhitelistedDir(directory) {
...
   return whitelist.indexOf(directory) > -1;
 }
+
+/**
+ * Tests if file is a Shopify theme settings file (settings_*.json)
+ *
+ * @param {string} directory - The path to the directory.
+ */
+export function isShopifyThemeSettingsFile(file) {
+  const whitelist = ['settings_schema.json', 'settings_data.json'];
```
t-kelly: These values are buried pretty deep in the code. Makes me think this package could use a config file. Definitely not needed in this PR though.

packages/slate-cli/src/commands/migrate.js
chrisberthe: Agreed.

packages/slate-cli/src/utils.js
```diff
@@ -178,3 +197,13 @@ export function isShopifyThemeWhitelistedDir(directory) {
...
   return whitelist.indexOf(directory) > -1;
 }
+
+/**
+ * Tests if file is a Shopify theme settings file (settings_*.json)
+ *
+ * @param {string} file - The path to the file.
+ */
+export function isShopifyThemeSettingsFile(file) {
+  const whitelist = ['settings_schema.json', 'settings_data.json'];
```
chrisberthe: Maybe for future proof, we can use a Regex to check whether or not the format is `settings_*.json` instead of having a hardcoded whitelist.

packages/slate-cli/src/commands/migrate.js
```diff
@@ -73,14 +74,29 @@ export default function(program) {
 
       const files = readdirSync(workingDirectory);
       const whitelistFiles = files.filter(isShopifyThemeWhitelistedDir);
-      const promises = whitelistFiles.map(movePromiseFactory);
+      const movePromises = whitelistFiles.map(movePromiseFactory);
+
+      function beautifyJsonPromiseFactory(file) {
+        console.log(`  Beautifying ${file}...`);
```
chrisberthe: See next comment.

packages/slate-cli/src/commands/migrate.js
maximevaillancourt: Ditto!

packages/slate-cli/src/commands/migrate.js
```diff
@@ -73,14 +74,29 @@ export default function(program) {
...
-        await Promise.all(promises);
+        await Promise.all(movePromises);
 
         console.log('');
         console.log(`  ${green(figures.tick)} Migration to src/ completed`);
         console.log('');
+
+        await Promise.all(beautifyPromises);
+
+        console.log(`  ${green(figures.tick)} Beautification of settings files completed`);
```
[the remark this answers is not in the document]
maximevaillancourt: Agreed.

packages/slate-cli/src/utils.js
maximevaillancourt: Could be as simple as that:
```
export function isShopifyThemeSettingsFile(file) {
  return /^settings_.+\.json/.test(file);
}
```

maximevaillancourt: Could be as simple as that:
```
export function isShopifyThemeSettingsFile(file) {
  return /^settings_.+\.json/.test(file);
}
```

packages/slate-cli/src/utils.js
```diff
@@ -140,6 +140,25 @@ export function hasDependency(dependencyName, pkg) {
   return hasDependencies;
 }
 
+
+/**
+ * Beautify a JSON file
+ *
+ * @param {string} file - The path to the file.
+ */
+export function beautifyJson(file) {
```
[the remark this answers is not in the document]
maximevaillancourt: I feel like `unminifyJson` assumes that the file is minified — which it might not be — though I don't hold to that opinion very strongly. `unminifyJson` works for me.

maximevaillancourt:
This PR is good for another round of review. @chrisberthe @t-kelly

packages/slate-cli/src/commands/migrate.js
```diff
@@ -4,7 +4,7 @@ import {join} from 'path';
 import {prompt} from 'inquirer';
 import {green, red, yellow} from 'chalk';
 import figures from 'figures';
-import {downloadFromUrl, startProcess, writePackageJsonSync, move, isShopifyTheme, isShopifyThemeWhitelistedDir} from '../utils';
+import * as utils from '../utils';
```
chrisberthe: Aww yeah.

PR #347, opened 2017-11-16
Add 'itemCondition' Schema.org meta tag http://schema.org/itemCondition

iamkeir:

Schema.org's `itemCondition` is an important (possibly mandatory) meta tag for products, proposing this be added to Slate. :tophat: @robcurry

http://schema.org/itemCondition

Issue: https://github.com/Shopify/slate/issues/348

### DIFF
--- a/packages/slate-theme/src/sections/featured-product.liquid
+++ b/packages/slate-theme/src/sections/featured-product.liquid
@@ -12,6 +12,7 @@
   <meta itemprop="url" content="{{ shop.url }}{{ product.url }}">
   <meta itemprop="image" content="{{ featured_image | img_url: '800x' }}">
   <meta itemprop="description" content="{{ product.description | strip_html | escape }}">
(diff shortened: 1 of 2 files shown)


jonathanmoore:
I see hardcoding the `itemCondition` to new as problematic.  We often see merchants using our themes selling products that are used or refurbished.  If anything, now that this is a required property by Google, I would like to see the addition of a condition selector in the Shopify admin so merchants can properly set products as new, used, refurbished, or damaged.  Then with the addition of a new liquid tag, we could pull that information into the theme.

iamkeir:
@jonathanmoore definitely agree. A placeholder/as new value would be a great addition in in the meantime, though. It could then be populated by Product Type on a custom basis perhaps.

NathanPJF:
It may be a good addition, but couldn't add it to Slate at this point because there is nothing in Shopify's admin that allows merchants to control its value.

I'll close this PR, but we can continue the discussion on the original issue #348

PR #355, opened 2017-11-27
Prefix mixin conditionals were not supporting 'all', fixed to allow

iamkeir:

The `prefix` mixin was not supporting 'all' due to the `if-else` conditionals. Switched to multiple `if` statements to fix.


iamkeir:
Hey @t-kelly - I thought I fixed a bug, but actually I think I added the 'all' prefix but had forgotten. Wasn't sure if it was therefore a worthy PR. I still use always make this change to my base Slate install so happy for it to be considered.

t-kelly:
It looks like this PR is just style changes. Closing.

Issue #243, opened 2017-08-22
Slate Watch Error: Transfer Failed: File(s) failed to upload to store. See log notes above.


ConduciveMammal:
### Problem

I've been in a situation where I've had to move my theme directory to and from BitBucket to pass work to colleagues and to work from home. After they've downloaded all of the files and attempted to start working, they use `slate watch` and it all starts fine, but as soon as they save a file, Slate CLI produces this error: `Transfer Failed: File(s) failed to upload to store. See log notes above.`. 
I looked in `deploy.log` but it seems it was empty. I figured it was a mismatch in differing Node/Slate versions. Now I'm on my home computer, downloaded from BitBucket, and I'm too now seeing this error whenever I try to push a new change to the store.

![image](https://user-images.githubusercontent.com/18140157/29590403-8d02e484-8791-11e7-980d-aeb6649cac30.png)

### More Information

Running `node -v` says my node version is v6.11.0
Running `npm -v` says my NPM version is 5.3.0
Running `slate -v` says my Slate version is 0.11.0

carlosepp:
facing the same issue

carlosepp:
fixed by wiping out `node_modules` and slate start again

iamkeir:
Having same issue - colleague pulled down repo, ran `npm install`, ran `slate start`, everything begins running fine. However, when saving a file, we receive `Transfer Failed: File(s) failed to upload to store. See log notes above.` ...but no notes above.

`node -v` v8.8.1
`npm -v` v5.4.2
`slate -v` v0.13.0

Does `package-lock.json`, `deploy.log` or `theme.lock` have any impact on this? If so I can post the contents of them. Quite a critical issue as we are struggling to get setup for working together.

ConduciveMammal:
I can’t remember how I got passed the issue, if I remember rightly, my colleague’s and I had different versions of some packages, try and update them all on both yours and their machines and see if that helps.

iamkeir:
@ConduciveMammal thanks - will explore this (nice username btw).

Weirdly, we managed to get my colleague up and running by deleting `node_modules`, changing to Node v6 then running `npm install`. Problem is, now **_I_** am the one with the issue!

Will report back.

iamkeir:
Right, it seems to be working now - not sure which thing did it but I did the following:

1. Delete `node_modules`, `package-lock.json`, `deploy.log` and `theme.lock` if you have them
2. Switch to Node `v6`
[...]

PR #366, opened 2017-12-12
Make sure we log errors on watch command

t-kelly:
Fixes https://github.com/Shopify/slate/issues/243

Its not possible to see error details if the error thrown by the Themekit child process is not a string. This makes sure that we log all watch errors in the deploy log, including those which are not strings.

### DIFF
--- a/packages/slate-tools/src/tasks/includes/messages.js
+++ b/packages/slate-tools/src/tasks/includes/messages.js
@@ -36,7 +36,7 @@ const messages = {
 
   logTransferFailed: (errMsg) => {
     gutil.log('Transfer Failed:',
-      gutil.colors.yellow(`${typeof errMsg === 'string' ? errMsg : 'File(s) failed to upload to store. See log notes above.'}`),
+      gutil.colors.yellow(`${typeof errMsg === 'string' ? errMsg : 'File(s) failed to upload to store. See log notes in deploy.log'}`),
     );
   },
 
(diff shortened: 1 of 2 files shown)


packages/slate-tools/src/tasks/includes/messages.js
```diff
@@ -64,6 +64,16 @@ const messages = {
     return timestamp + action + amount + fileList;
   },
 
+  logDeployErrors: (cmd, files, err) => {
```
NathanPJF: Most of this code is the same as `logDeploys()`.  What about modifying that existing function to check for the presence of an `err`?

```js
logDeploys: (cmd, files, err) {
  const timestamp = `Deploy ${err ? 'error' : 'done'} @ ${new Date()}. `;
  const action = cmd === 'upload' ? 'added/changed ' : 'removed ';
  const amount = `${files.length} file(s): `;
  const fileList = `${files.join(', ')}.\n`;

  let logMessage = timestamp + action + amount + fileList;

  if (err) { logMessage += `${err} \n`}
  
  return logMessage;
}
```

tgdev:
@NathanPJF @t-kelly 

Any idea on if/when this will get merged in? Struggling with Cli telling me "File changes successfully synced to store" when updating `.scss` files but seeing broken local page.

Do you know of a temp work around to get an idea of being alerted to any possible sass errors in the meantime?

PR #388, opened 2018-01-26
Add support for type array in slate config

mmorissette:
Add support for type `array` in the `slate-config` package. Also added an `extractType` function so that we can easily extend the types supported by `slate-config`

### How is this accomplished
Add specific check for type `array` in `slate-config` schema validation.

### Checklist

For maintainers:


mmorissette:
@t-kelly I added a test to check that schema values of type `array` are validated properly. Can I get a second look from you 🙏

PR #413, opened 2018-02-26
Slate tools refactor

t-kelly:

Refactor `@shopify/slate-tools` package so that it will be easy to add additional tools like ESLint, Stylelint, and Prettier.

Changes `slate.config.js` to be namespaced by their package, for example:
```js
{
  cssVarLoaderEnable: false
}
```
is now
```js
{
  slateCssVarLoader: {
    cssVarLoaderEnable: false
  }
}
```

@mmorissette mind checking if the changes to `@shopify/slate-config` don't break CSS Var Loader?


t-kelly:
@mmorissette I realized I didn't have slate-tools properly linked when testing 😅 

Pushed some fixes. Mind giving it another shot? Your webpack config doesn't need to change.

Will work on adding some tests now.

t-kelly:
Webpack configs have been moved to:
```
{
  slateTools: {
    extends: {
      dev: { ... },
      prod: { ... }
    }
}
```

packages/slate-tools/tools/webpack/get-user-webpack-config.js
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable global-require, import/no-dynamic-require */
-const {getSlateConfig} = require('@shopify/slate-config');
+const config = require('../../slate-tools.config');
```
mmorissette: This is loading the package schema and not the actual user config. Because of this, when trying to build a theme, it does not consider my config and therefore does not run my webpack extends.

t-kelly: Really need to add tests to this package...

t-kelly: This is hooked up correctly. What I forgot is to add `extends` to the schema.

t-kelly:
@mmorissette okay `extend` values in `slate.config.js` SHOULD be applied now.

packages/slate-tools/tools/webpack/get-user-webpack-config.js
mmorissette: 🤦‍♂️ my bad @t-kelly, you are correct.

PR #446, opened 2018-03-20
Fix Node version check and error

t-kelly:
Previously, the JS file that checked what version of Node was running was using JS language features available only in Node 6+. This caused a runtime error when the file was run in Node <6.

Switch out Node 6+ features for equivalent legacy alternatives so the Node version warning can be displayed. Disable ESLint so that Prettier does not apply ES6 formatting.


NathanPJF:
Thanks @t-kelly - sorry I was slow on the review 😅

Issue #585, opened 2018-05-23
Fill in analytics email with an .env variable


markdavies:
### Problem

Sometimes we like to run the slate deploy commands from a deployment service (ie run a script and not have human input). In v1 this means that deployment halts when the analytics question is asked.

It would be great if we could use an .env variable (or pass a value somehow) to auto-answer this question?

t-kelly:
That is a brilliant idea!

Issue #552, opened 2018-05-14
User cannot opt-out of Slate Analytics


madma:
### Problem

User is not offered the option of opting out of Slate Analytics, nor does analytics "opt-out" seem to be a feature that has been built-in yet hidden at this time, after inspecting relevant source code: slate/packages/slate-analytics/index.js (last commit affecting this file: https://github.com/Shopify/slate/commit/e8b5b1d590a2052daa891ed22fbf2606cf7be011).

This is misleading, as the documentation suggests that a user is able to opt-out of analytics: https://github.com/Shopify/slate/wiki/Slate-Analytics (last commit: 7ebceb3).

### Replication steps

1. Run `yarn create slate-theme my-new-theme`
2. User is prompted for email to provide consent for analytics.
3. There is no option for user to use Slate without consenting to analytics.
4. On Mac: `cmd + c` to interrupt the theme scaffold script.
5. An empty theme folder is generated.

t-kelly:
@madma This was a choice the team made for pre-releases of Slate v1. We intend to make analytics optional when we release the final version of Slate 1.0.0.

PR #592, opened 2018-05-24
Add SLATE_USER_EMAIL env variable for slate-analytics

t-kelly:

Fixes https://github.com/Shopify/slate/issues/585
Related https://github.com/Shopify/slate/issues/552

Adds a new environment variable `SLATE_USER_EMAIL` to `@shopify/slate-env`. When setting this value to a valid email, `@shopify/slate-analytics` will use it to consent to data collection instead of prompting the user.

cc @markdavies

### DIFF
--- a/packages/slate-analytics/__tests__/index.test.js
+++ b/packages/slate-analytics/__tests__/index.test.js
@@ -1,3 +1,5 @@
+/* eslint-disable no-process-env */
+
 jest.mock('../prompt');
 jest.mock('axios');
 
@@ -98,6 +100,24 @@ describe('init()', () => {
       expect(config.tracking).toBe(true);
       expect(rc.get().tracking).toBe(true);
       expect(config.trackingVersion).toBe(packageJson.trackingVersion);
+    });
+  });
+
+  describe('skips prompting new user for tracking consent', () => {
+    test('if SLATE_USER_EMAIL environment variable is set to a valid email', async () => {
+      const analytics = require('../index');
+      const prompt = require('../prompt');
+      const mock = require('mock-fs');
+
+      process.env.SLATE_USER_EMAIL = 'test@email.com';
+
+      mock();
+
+      await analytics.init();
+
+      expect(prompt.forNewConsent).not.toHaveBeenCalled();
+
+      delete process.env.SLATE_USER_EMAIL;
     });
   });
 });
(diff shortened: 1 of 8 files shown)


packages/slate-analytics/prompt.js
```diff
@@ -2,16 +2,14 @@ const inquirer = require('inquirer');
...
 const question = {
   type: 'input',
   name: 'email',
   message: 'To continue, please enter your email address:',
   validate: (input) => {
-    const re = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
-    return re.test(String(input).toLowerCase())
-      ? true
-      : 'Email not valid. Please try again.';
+    return validateEmail(input) ? true : 'Email not valid. Please try again.';
```
chrisberthe: Same thing but could do this instead:
```js
return validateEmail(input) || 'Email not valid. Please try again.';
```

packages/slate-env/__tests__/index.test.js
```diff
@@ -14,6 +14,7 @@ const TEST_ENV = {
   [config.envPasswordVar]: '123456789',
   [config.envThemeIdVar]: '987654321',
   [config.envIgnoreFilesVar]: 'config/settings_data.json',
+  [config.envUserEmail]: 'thomas.kelly@shopify.com',
```
chrisberthe: Can probably keep it consistent and use `test@email.com` like you're doing above in `packages/slate-analytics/__tests__/index.test.js`.

packages/slate-analytics/__tests__/index.test.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-process-env */
```
chrisberthe: This is being referenced a few times throughout the codebase - we could probably add this to the list of rules in `.eslintrc`.

t-kelly: Don't want to apply this rule globally. In general I feel like this rule is helpful -- it makes you question you reallllly need to use process-env or not.

chrisberthe: Fair enough.

Issue #575, opened 2018-05-20
Loading Webpack Plugins to expose jQuery to 3rd party modules


lmartins:
### Problem
I am trying to use https://github.com/discolabs/cartjs as a CommonJS module, which is a library that depends on `jQuery` to be available on the window object.

One way Webpack allows to use this is via the `ProvidePlugin` https://webpack.js.org/plugins/provide-plugin/, so I wonder how can we achieve that using Slate's config file.

### More Information

I've tried to expose a plugins array to the dev config, but that unfortunately didn't worked:
```js

const path = require('path');

const alias = {
  jQuery: path.resolve('./node_modules/jquery'),
  jquery: path.resolve('./node_modules/jquery'),
  'lodash-es': path.resolve('./node_modules/lodash-es')
};

const plugins = [
  new webpack.ProvidePlugin({
    jQuery: 'jquery',
    $: 'jquery',
    jquery: 'jquery'
  })
];

module.exports = {
  slateCssVarLoader: {
    cssVarLoaderLiquidPath: ['src/snippets/css-variables.liquid']
  },
  slateTools: {
    extends: {
      dev: { resolve: { alias }, plugins: plugins },
      prod: { resolve: { alias } }
    }
  }
};
```

The only thing that worked was to create a jQuery global variable inside my module, which feels like a wrong approach to it. 

Any tips on better solution to this?

idpocky:
I am trying to do this. Because I got the message `WARNING in entrypoint size limit...` while slate building.

1. Include jQuery as a 3rd by the `externals` of webpack in 'production' to reduce the size of bundle file. It can be done by the following code.
Refer: [webpack externals](https://webpack.js.org/configuration/externals/)

```javascript
// ./slate.config.js
const externals = {
  jquery: 'jQuery',
};

module.exports = {
  slateCssVarLoader: {
    cssVarLoaderLiquidPath: ['src/snippets/css-variables.liquid'],
  },
  slateTools: {
    extends: {
      dev: {
        resolve: {alias},
      },
      prod: {
        resolve: {alias},
        externals,
      },
    },
  },
};
```
```html
<!-- ./src/layout/theme.liquid -->
<head>
...
<script 
      src="https://code.jquery.com/jquery-3.3.1.min.js"
      integrity="sha256-FgpCb/KJQlLNfOu91ta32o/NMZxltwRo8QtmkMRdAu8="
      crossorigin="anonymous"></script>
{% include 'script-tags', layout: 'theme' %}
...
[...]

PR #598, opened 2018-05-26
Add webpack.ProvidePlugin config object

jonathanmoore:

Add a way to pass through a [webpack.ProvidePlugin](https://webpack.js.org/plugins/provide-plugin/) object from the `slate.config.js` file to the `core.js` configuration.  Originally, I had attempted to add the webpack.ProvidePlugin plugin to the dev/prod extends in `slate.config.js`, however I'm fairly certain that it was not working because it was loaded prior to the HotModuleReplacementPlugin.

- Issue: Loading Webpack Plugins to expose jQuery to 3rd party modules #575

Below is an example `slate.config.js` file that automatically loads $ and jQuery modules, which solves issues with including most jQuery plugins where jQuery would not be defined.  Additionally, there is another example where Lodash Map is added as a provide plugin.

```js
/* eslint-disable no-undef */

const path = require('path');

const alias = {
  jquery: path.resolve('./node_modules/jquery'),
  'lodash-es': path.resolve('./node_modules/lodash-es'),
};

module.exports = {
  slateCssVarLoader: {
    cssVarLoaderLiquidPath: ['src/snippets/css-variables.liquid'],
  },
  slateTools: {
    providePlugin: {
      $: 'jquery',
      jQuery: 'jquery',
      _map: ['lodash-es', 'map'],
    },
    extends: {
      dev: {resolve: {alias}},
      prod: {resolve: {alias}},
    },
  },
};
```

### Checklist
For contributors:

For maintainers:

As an alternate approach, it might make sense to add a way to extend core for plugins or loaders that need to be covered across dev/prod or need to be added prior to dev/prod.

### DIFF
--- a/packages/slate-tools/slate-tools.config.js
+++ b/packages/slate-tools/slate-tools.config.js
@@ -36,6 +36,13 @@ module.exports = generate({
       default: true,
     },
     {
+      id: 'providerPlugin',
+      default: {},
+      description:
+        'Automatically load modules instead of having to import or require them everywhere',
+      type: 'object',
+    },
+    {
       id: 'extends',
       items: [
         {
--- a/packages/slate-tools/slate-tools.config.js
+++ b/packages/slate-tools/slate-tools.config.js
@@ -36,7 +36,7 @@ module.exports = generate({
       default: true,
     },
     {
-      id: 'providerPlugin',
+      id: 'providePlugin',
       default: {},
       description:
         'Automatically load modules instead of having to import or require them everywhere',
--- a/packages/slate-tools/tools/webpack/config/core.js
+++ b/packages/slate-tools/tools/webpack/config/core.js
@@ -133,7 +133,7 @@ module.exports = {
   plugins: [
     ...contextReplacementPlugins(),
 
-    new webpack.ProvidePlugin(config.providerPlugin),
+    new webpack.ProvidePlugin(config.providePlugin),
 
     extractLiquidStyles,
 
--- a/packages/slate-tools/tools/webpack/config/core.js
+++ b/packages/slate-tools/tools/webpack/config/core.js
@@ -133,6 +133,8 @@ module.exports = {
   plugins: [
     ...contextReplacementPlugins(),
 
+    new webpack.ProvidePlugin(config.providerPlugin),
+
     extractLiquidStyles,
 
     new CopyWebpackPlugin([


packages/slate-tools/slate-tools.config.js
```diff
@@ -35,6 +35,13 @@ module.exports = generate({
         'Enable/disable the prompt to skip uploading settings_data.json',
       default: true,
     },
+    {
+      id: 'providerPlugin',
```
chrisberthe: Might be a typo but we should keep this consistent with the `providePlugin` plugin name.

jonathanmoore: Good catch! I think "provider" somehow got stuck in my head and I ran with it for everything—including this PR.

t-kelly:
@jonathanmoore does this plugin solve the problem of making jQuery available globally outside of the JS bundles, as mentioned in #575? I think this just makes`jquery` and/or `$` available on every JS file.

I'm not sure if I'm personally a fan of this pattern. It's returning one of the pitfalls of global variables that you never really know what is assigned to it. The benefit of calling `import $ from 'jquery'` is that any developer can walk into a JS file and know the full context of the dependencies they are working with...

I think I would prefer to get this working with the extend feature rather then a dedicated integration.

jonathanmoore:
I personally don't see a way around this issue in the near future.  The webpack ProvidePlugin was created for this specific use case in mind https://webpack.js.org/plugins/provide-plugin/

There was a plugin that would allow for resource by resource injection of global objects, however it's stuck at support for webpack 1.0 https://github.com/webpack-contrib/imports-loader

The reality is a lot of themes still need to utilized jQuery plugins for complex functionality where it doesn't always make sense to rewrite it from scratch.  It's great that Shopify is moving away from jQuery, but we're not ready to make that switch fully as well.

Then we also run into the challenge of properly associating the correct version of jQuery with the theme.  In real world merchant stores we often see issues where Shopify Apps will load a wide variety of jQuery versions, and occasionally do it in a way that overwrites and kills the theme's version we depend on.  If we are going to use a package like jQuery we need a reliable way to ensure it's isolated and used properly for the theme.

jonathanmoore:
You're right @t-kelly, I should have dug more to look into ways to include globals only for legacy modules.  The first round of research over the weekend pointed to ProvidePlugin as the solution.

However, the correct approach is outlined on Webpack's docs under Granular Shimming.
https://webpack.js.org/guides/shimming/#granular-shimming
> The webpack compiler can understand modules written as ES2015 modules, CommonJS or AMD. However, some third party libraries may expect global dependencies (e.g. $ for jQuery). The libraries might also create globals which need to be exported. These "broken modules" are one instance where shimming comes into play.

Here is a solution that does not require this PR or ProvidePlugin.  In this example the legacy `@fancyapps/fancybox` module has an expectation for a global jQuery variable.  For demonstration purposes I'm also adding in an assignment of `$=jquery`

Add [imports loader plugin](https://webpack.js.org/loaders/imports-loader/)
`yarn add imports-loader`

```js
// slate.config.js

/* eslint-disable no-undef */

const path = require('path');

const alias = {
  jquery: path.resolve('./node_modules/jquery'),
  'lodash-es': path.resolve('./node_modules/lodash-es'),
};

module.exports = {
  slateCssVarLoader: {
    cssVarLoaderLiquidPath: ['src/snippets/css-variables.liquid'],
  },
  slateTools: {
    extends: {
      dev: {
        resolve: {alias},
        module: {
          rules: [
            {
              test: require.resolve('@fancyapps/fancybox'),
              use: 'imports-loader?$=jquery,jQuery=jquery',
            },
          ],
        },
      },
      dev: {
        resolve: {alias},
        module: {
          rules: [
            {
              test: require.resolve('@fancyapps/fancybox'),
              use: 'imports-loader?$=jquery,jQuery=jquery',
            },
          ],
        },
      },
    },
  },
};
```

PR #532, opened 2018-05-04
Modified slate-cssvar-loader to allow CSS variables without Liquid variables

truribe:
Modified slate-cssvar-loader not to prevent deploying when no matching liquid variable is found for a CSS variable.

### What are you trying to accomplish with this PR?

I'm attempting to solve the issue I reported that prevents the use of regular CSS variables without corresponding Liquid variables.
https://github.com/Shopify/slate/issues/524

I do not believe any documentation for local SASS compilation needs to be updated to reflect this change. This documentation is the most relevant:
https://github.com/Shopify/slate/wiki/Local-SASS-compilation

### Checklist
For contributors:
- [x] I have [updated the docs](https://github.com/Shopify/slate/blob/master/CONTRIBUTING.md#documentation) to reflect these changes, if applicable.

For maintainers:
- [ ] I have :tophat:'d these changes.

PR #619, opened 2018-06-04
Revert "Modified slate-cssvar-loader to allow CSS variables without Liquid variables"

t-kelly:
Reverts Shopify/slate#532

@truribe your previous PR broke some of our tests. Could you reopen it and double check the tests that fail?


truribe:
In #532 I neglected to remove the test that expected the exception when I removed the exception. PR #621 removes the exception and the offending test.

PR #643, opened 2018-06-20
Throw an error if we fail to access Theme API

### LINKED ISSUE #428: `npm start` fails silently if the API is set to NO ACCESS
### Problem
While developing a theme:

Running `npm start` fails silently if the user hasn't granted access:

image

We should show an error stating that the user needs to grant access.

### Replication steps

- Set API access to `no access`
- Run `npm start`

@t-kelly

t-kelly:

Fixes #428 

We were not catching any request errors when checking if we were deploying to the store's published theme. This PR makes sure we are handling any errors emitted from this action.

### DIFF
--- a/packages/slate-sync/index.js
+++ b/packages/slate-sync/index.js
@@ -1,5 +1,6 @@
 const chalk = require('chalk');
 const figures = require('figures');
+const https = require('https');
 const themekit = require('@shopify/themekit').command;
 const slateEnv = require('@shopify/slate-env');
 const config = require('./slate-sync.config');
@@ -66,7 +67,7 @@ async function deploy(cmd = '', files = []) {
     await promiseThemekitConfig();
     await promiseThemekitDeploy(cmd, files);
   } catch (error) {
-    console.error(error);
+    console.error('My Error', error);
   }
 
   deploying = false;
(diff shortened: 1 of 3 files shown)


packages/slate-tools/cli/commands/start.js
```diff
@@ -23,7 +23,7 @@ const options = {
 const spinner = ora(chalk.magenta(' Compiling...'));
 let firstSync = true;
 let skipSettingsData = null;
-let continueIfPublishedTheme = null;
+const continueIfPublishedTheme = null;
```
chrisberthe: Should be `let`.

packages/slate-tools/cli/commands/start.js
```diff
@@ -81,13 +81,22 @@ devServer.compiler.hooks.done.tap('CLI', (stats) => {
 });
 
 devServer.client.hooks.beforeSync.tapPromise('CLI', async (files) => {
+  console.log('sync');
```
chrisberthe: We can remove this.

packages/slate-sync/index.js
```diff
@@ -52,6 +53,7 @@ function _generateIgnoreFlags() {
  * @return          Promise
  */
 async function deploy(cmd = '', files = []) {
+  console.log('deploy start');
```
chrisberthe: We can remove this.

packages/slate-sync/index.js
```diff
@@ -119,6 +121,70 @@ function promiseThemekitDeploy(cmd, files) {
...
+          if (!Array.isArray(parsed.themes)) {
+            reject(`
+              Shopify response for /admin/themes.json is not an array.
+
+              ${JSON.stringify(parsed, null, '\t')}
+            `);
+            return;
+          }
+
+          const mainTheme = parsed.themes.find((t) => t.role === 'main');
```
chrisberthe: Out of curiosity, are there ever any instances where the published theme doesn't have a role of "main"?

t-kelly: Hmmm... @davidcornu? 👋^^

NathanPJF: @chrisberthe yep, for legacy reasons a published theme can have a role of `main` or `mobile`.  There was once a time when Shopify supported two published themes - one for desktop and one for mobile. Mobile has long been deprecated, but may exist in the wild.  So you want to make sure you're pushing to the `main` one always.

chrisberthe: Interesting! Makes sense, thanks for the clarification.

Issue #611, opened 2018-05-30
Assets/images skipped on build or deploy


jonathanmoore:
### Problem

Images placed into `/src/assets/images` are not copied into `/dist/assets` on build or deploy.

### Replication steps

1. Place an image (jpeg, jpg, svg, png, gif) into `/src/assets/images`
2. Run `yarn build` or `yarn deploy`
3. The expectation would be to find the images in `/dist/assets` but nothing is copied over

### More Information

It looks like there is a missing paths item in `slate-tools.config.js` for `images` that is referenced in `core.js`. I tried locally fix it hoping it was a simple fix, but I think the issue is bigger than just a path reference.

https://github.com/Shopify/slate/blob/d67f445ea4dda555cce836f58d78f23ec3b55b84/packages/slate-tools/slate-tools.config.js#L73-L99
https://github.com/Shopify/slate/blob/d67f445ea4dda555cce836f58d78f23ec3b55b84/packages/slate-tools/tools/webpack/config/core.js#L47-L54

wrozka:
Any updates on this one? I migrated my 0.x template to 1.x but I can't add any new images now.

petema3:
FWIW I also was having trouble with this until I realized that it only uploads images from `/src/assets/images/` if:

1. It's actually used in your theme somewhere.
2. The local path is used, with single quotes around the file path, not double quotes.

So it didn't upload the images when I didn't have the image referenced in the theme yet anywhere, or if I used the old way with just the filename like this `{{ "placeholder.png" | asset_url }}`.  But if I add it in the theme as `{{ '../assets/images/placeholder.png' | asset_url }}` it works.

I'm not sure if this is a bug or a feature, but this is the functionality as I've mapped it out through trial and error.

xavianaxw:
#557 and #559 talks about this

DPerkunas:
Any news on this? Same issue, using {{ '../assets/images/placeholder.png' | asset_url }} with asset in /src/assets/images/placeholder.png didn't work for me like it did for petema3. 

A url is generated by shopify, but no images found uploaded into shopify's theme/assets folder.

xavianaxw:
@DPerkunas have you tried `yarn deploy`

DPerkunas:
@xavianaxw Yeah I did. Maybe the issue is the fact that I'm deploying to a live site but here's the full scenario of what I noticed.

While having yarn watch running.

[...]

PR #647, opened 2018-06-22
Copy the contents of src/assets/images and src/assets/fonts

t-kelly:

Fixes https://github.com/Shopify/slate/issues/611

Previously, to include an image in the theme build, it needed to be referenced somewhere in your theme's liquid:

```
<img src="{{ '../assets/images/placeholder.png' | asset_url }}">
```

This is because Slate needs to transform urls made with `| asset_url` to a localhost equivalent for development. For example, the above Liquid gets transformed to the following when running `slate-tools start`:

```
<img src="https://localhost:8080/placeholder.png">
```

The above approach is still recommended in order to properly load images while developing because it's not guaranteed that the image will be uploaded to Shopify CDNs. `slate-tools start` only uploads Liquid and JSON files to Shopify for development and relies on the local asset server to serve up all other assets from localhost.

However, an image may still be needed even if its not referenced in a Theme's liquid, for example if its generated dynamically after pageload. Its for for this reason we should copy all images to the `dist/assets` folder.

Please note that for `slate-tools start` images will still be served from localhost, so you'll need to make sure any URL you are generating is pointing to there.

### DIFF
--- a/packages/slate-tools/slate-tools.config.js
+++ b/packages/slate-tools/slate-tools.config.js
@@ -87,7 +87,16 @@ module.exports = generate({
         },
         {
           id: 'static',
-          default: resolveTheme('src/assets/static'),
+          items: [
+            {
+              id: 'src',
+              default: resolveTheme('src/assets/static'),
+            },
+            {
+              id: 'dist',
+              default: resolveTheme('dist/assets'),
+            },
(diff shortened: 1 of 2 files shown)


yankeyhotel:
Should this work in scss files?

Issue #657, opened 2018-06-29
Yarn deploy fails 422 "theme_packaged_with" is not a valid attribute


MitchEff:
### Problem
'yarn start' and 'yarn build' are working perfectly fine. All of a sudden, using 'yarn deploy' fails, saying "theme_packaged_with" is not a valid attribute. 

I can see in 'dist/config/settings_schema.json', it's adding this just sort of randomly in the middle of the file: 

""theme_packaged_with":"@shopify/slate-tools@1.0.0-beta.1"

Screenshot [here](https://imgur.com/a/EN6d2Mw). 

### Replication steps
Use 'yarn deploy', then go check your dist/config/settings_schema.json. It _might_ be there? Just appeared today for us.

Skoulix:
Same thing happens to me. Deploy command fails.

t-kelly:
@MitchEff was this working before? This attribute should be whitelisted on Shopify servers...

MitchEff:
Yeah, it was working okay a couple days prior. 

Oddly enough, with beta 5 it fails and then doesn't deploy, but on beta 1 it fails but still deploys everything except settings schema, which I'm just copying manually for now.

t-kelly:
Looking into it! 🕵️

Skoulix:
I don't have the warning any more so this is indeed whitelisted by Shopify servers. Honestly I had the warning for a couple of days, everything seems fine now. 

Thank you.

t-kelly:
I can confirm that this issue occurs when the object with `name: "theme_info"` is not the first object in the `settings_schema.json` array.

PR #662, opened 2018-07-03
Only set theme_packaged_with if the first settings panel is theme_info

davidcornu:

Fixes https://github.com/Shopify/slate/issues/657.

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/packages/slate-tag-webpack-plugin/index.js
+++ b/packages/slate-tag-webpack-plugin/index.js
@@ -1,4 +1,5 @@
 const SETTINGS_SCHEMA_PATH = '../config/settings_schema.json';
+const THEME_INFO_PANEL = 'theme_info';
 
 function FileListPlugin(version) {
   this.version = version;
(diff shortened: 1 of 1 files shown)


packages/slate-tag-webpack-plugin/index.js
```diff
@@ -9,7 +10,11 @@ FileListPlugin.prototype.apply = function(compiler) {
     const asset = compilation.assets[SETTINGS_SCHEMA_PATH].source();
     const schema = JSON.parse(asset);
 
-    if (Array.isArray(schema) && typeof schema[0] === 'object') {
+    if (
+      Array.isArray(schema) &&
+      typeof schema[0] === 'object' &&
+      schema[0].name === THEME_INFO_PANEL
```
t-kelly: Instead of needing it to be first settings panel, can we just find the schema item with `name == THEME_INFO_PANEL`?

davidcornu: @t-kelly at first I thought `theme_info` had to be first but double-checked and looks like it can be anywhere. Will update the PR.

davidcornu: Done.

PR #674, opened 2018-07-13
Fix middleware module overwriting compiler watchOptions

### LINKED ISSUE #672: Unable to extend webpackDevMiddleware options
### Problem

I need to enable polling in the webpackDevMiddleware `watchOptions` object for watching to work because we're running a VM through VirtualBox and using NFS. I have tried extending the `dev` configuration in `slate.config.js` like so:

    slateTools: {
        extends: {
            dev: {
                resolve: {alias},
                watchOptions: {
                    poll: true,
                    ignored: /node_modules/
                }
            },
            prod: {resolve: {alias}},
        }
    }

But this doesn't seem to have an effect. I make changes to files and nothing gets triggered. 

By editing `node_modules/@shopify/slate-tools/tools/dev-server/app.js` as below the file changes are seen and everything works correctly:

    app.webpackDevMiddleware = webpackDevMiddleware(compiler, {
      quiet: true,
      reload: true,
      watchOptions: {
        poll: true,
        ignored: /node_modules/
      }
    });

Could you add something in so that the options object passed to `webpackDevMiddleware` can be extended in `slate.config.js`?

andyexeter:
The `webpack-dev-middleware` module sets the compiler's `watchOptions` object even if the middleware module's object is empty and the compiler's is not. This means any changes made to the compiler config `watchOptions` object are overwritten and not used.

Fixed by passing in the compiler's `watchOptions` object if it is set.

I can't submit a PR to fix the issue in the `webpack-dev-middleware` module because slate-tools requires v1.10.2 which is two major versions behind.

Fixes #672

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/packages/slate-tools/tools/dev-server/app.js
+++ b/packages/slate-tools/tools/dev-server/app.js
@@ -10,7 +10,7 @@ module.exports = class App {
     app.webpackDevMiddleware = webpackDevMiddleware(compiler, {
       quiet: true,
       reload: true,
-      watchOptions: compiler.options.watchOptions || {}
+      watchOptions: compiler.options.watchOptions || {},
     });
     app.webpackHotMiddleware = webpackHotMiddleware(compiler, {
       log: false,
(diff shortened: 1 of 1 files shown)


t-kelly:
Hey @andyexeter, shall we upgrade to the latest version while we're at it?

t-kelly:
Closing due to lack of response. From the sounds of it, this should be fixed in `webpack-dev-middleware`?

PR #681, opened 2018-07-18
Add an option to disable SSL.

ajw0100:
At the moment, I can't seem to get Slate's local Express server to work with AWS Cloud9. This is a small, quick and dirty PR that fixes this. 

C9 only connects to local apps via HTTP (as documented here: https://docs.aws.amazon.com/cloud9/latest/user-guide/app-preview.html#app-preview-run-app). However, it will serve these apps to remote browsers over HTTPS. So simply disabling SSL allows Slate to be used from a Cloud9 instance.

I don't code much in JS and I'm not familiar with all the JS dev tools so please forgive the lack of tests and any other sins. I've tested it with my C9 instance and a fresh Shopify store and it works like a charm. Is this something the Slate team is willing to officially support?

### DIFF
--- a/packages/slate-tools/slate-tools.config.js
+++ b/packages/slate-tools/slate-tools.config.js
@@ -49,6 +49,10 @@ module.exports = generate({
       ],
     },
     {
+      id: 'ssl',
+      default: true,
+    },
+    {
       id: 'domain',
       default: 'https://localhost',
     },
--- a/packages/slate-tools/tools/dev-server/index.js
+++ b/packages/slate-tools/tools/dev-server/index.js
@@ -1,6 +1,7 @@
 const fs = require('fs');
 const webpack = require('webpack');
-const {createServer} = require('https');
+const {createServer: createHttpServer} = require('http');
+const {createServer: createHttpsServer} = require('https');
 const createHash = require('crypto').createHash;
 
 const App = require('./app');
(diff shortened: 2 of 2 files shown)


andyexeter:
This would be very useful. Currently having to explain to any of our front end team that they need to visit the local URL in their browser first and add an SSL exception and it's confusing them.

jonathanmoore:
@andmcka82 Although if you disable SSL you're going to run into all sorts of security issues where browsers will fail to load insecure (http) resources while loading the secure (https) Shopify site.

Adding the trusted SSL certificate is a one-time fix that will prevent developers from having to manually click the exception each time.
https://github.com/Shopify/slate/wiki/How-to-create-a-trusted-local-SSL-certificate

ajw0100:
Yes you definitely want your local resources served over HTTPS to avoid mixed content warnings. However, in my case AWS Cloud9 takes care of serving the resources over HTTPS but it can only access those resources if they are served locally on the Cloud9 instance over HTTP.

andyexeter:
@jonathanmoore Good point, I don't know why that didn't cross my mind!

t-kelly:
@ajw0100 would you mind elaborating on how you're using AWS Cloud9? Interested to know how you're using Slate's dev pipeline.

Sounds like a reasonable request. Having a look through the code now.

t-kelly:
I've long been unhappy with Slate's current implementation of its config -- thus I haven't moved forward with documenting it properly and expanding it. I've been refactoring how we approach config with Slate (mostly internal) and it will affect this PR.

I will keep this PR open come the time I am able to look at it closer and potentially do a follow PR for it to work with the new config setup. Will keep you posted!

t-kelly:
@ajw0100 I'm going to close this because it the use case is very limited to your particular application. In all other cases, disabling SSL will result in a security errors. If this feature becomes more demanded I will consider adding it. Until then, I recommend running a forked version of Slate.

Issue #540, opened 2018-05-09
Add PostCSS to slate.config.js


RustyDev:
Is there any documentation for adding a PostCSS plugin in the build process? I'd like to use [Tailwind CSS](https://tailwindcss.com/docs/installation#using-tailwind-with-postcss).

t-kelly:
I am using PostCSS in the webpack build. I can make it configurable via `slate.config.js`. I'm going to change the title of this and use this issue to keep track of the progress.

anjolovic:
@t-kelly Any progress on the use of PostCSS plugins? I'd love to be able to start using TailwindCSS soon.

mrobit:
Sorry to be a bother - any update on this? @t-kelly you'd mentioned that your'e using PostCSS in the webpack build - can you explain a little more on how that works? I saw your [comment about slate.config.js and webpack configs here](https://github.com/Shopify/slate/issues/394#issuecomment-374303889)  so I wasn't sure if that's where I should start or not.

seandogg:
+1 -- looking to add some CSS Grid support in IE11 and add some PostCSS plugins to the build.

dan-gamble:
I don't think that it quite fits with this issue but i've implemented PostCSS using slate.config.js here: https://github.com/Shopify/slate/issues/670

It does basically replace Sass with PostCSS so it's not really using PostCSS in addition to Sass but i thought it may be helpful.

dan-gamble:
I'm sure there's more to it than what i've done here but i've started a PR that could get a decent initial version of this going: https://github.com/Shopify/slate/pull/683

pveen2:
@mrobit Did you already manage to get the tailwind x Slate working, i want to start with this combination but only if it isn't a big hassle?

PR #683, opened 2018-07-18
Add the ability to pass your own PostCSS plugins

dan-gamble:

https://github.com/Shopify/slate/issues/540

This will allow the ability to pass Postcss plugins to the postcss-loader in the Webpack config.

This is a very simple version to start with, it could be improved by moving the Autoprefixer & CSSNano parts in to the `slate.config.js` so the user can control the order of them.

I opted to put the custom plugins before the Autoprefixer and CSSNano plugins as i think 99.2% of use cases will want their plugins before any of these so for the initial version it would make sense.

Happy to be pointed towards a relevant place to detail this change somewhere for people.

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/packages/slate-tools/slate-tools.config.js
+++ b/packages/slate-tools/slate-tools.config.js
@@ -28,13 +28,13 @@ module.exports = generate({
       items: [
         {
           id: 'dev',
-          default: []
+          default: [],
         },
         {
           id: 'prod',
-          default: []
-        }
-      ]
+          default: [],
+        },
+      ],
     },
     {
       id: 'cssnanoSettings',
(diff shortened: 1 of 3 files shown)


t-kelly:
I've long been unhappy with Slate's current implementation of its config -- thus I haven't moved forward with documenting it properly, expanding it, and reviewing PRs on it. I've been refactoring how we approach config with Slate (mostly internal) and it will affect this PR.

I will keep this PR open come the time I am able to look at it closer and potentially do a follow PR for it to work with the new config setup. Will keep you posted!

dan-gamble:
Sounds good to me @t-kelly, looking forward to see what you come up with.

t-kelly:
Replaced by https://github.com/Shopify/slate/pull/753

PR #730, opened 2018-09-06
Correct tests to use global.slateUserConfig

t-kelly:

Followup to #729 and #725 

- Simplify how @shopify/slate-config uses global.slateUserConfig
- Remove uneeded, undocumented logic from @shopify/slate-config


dan-gamble:
You're on fire Thomas 👏

t-kelly:
YOLO 🚢'ing 😅

dan-gamble:
Loving it 😍

PR #780, opened 2018-09-26
Add config to enable/disable sourcemaps for styles

### LINKED ISSUE #447: CSS Sourcemaps activation in dev mode (1.x)
### Problem

It would be useful to be able to activate CSS/SCSS sourcemaps in dev mode.

If I go and edit directly *webpack.dev.conf.js* inside the node_modules the sourcemaps are working fine, but this is only a temporary solution.

If I understand correctly in #394 there is mention that by adding a config object inside _slate.config.js_ it should override the config inside _node_modules_, but it doesn't seem to work :

### Code inside _slate.config.js_

```javascript
/* eslint-disable no-undef */

const path = require('path');

const alias = {
  jquery: path.resolve('./node_modules/jquery'),
  'lodash-es': path.resolve('./node_modules/lodash-es'),
};

module.exports = {
  extends: {
    dev: {
        resolve: {alias},
        module: {
            rules: [
              {
                test: /\.s[ac]ss$/,
                exclude: commonExcludes(),
                use: [
                  {
                    loader: 'style-loader',
                    options: {
                      hmr: true,
                    },
                  },
                  {
                    loader: 'css-loader',
                    // Enabling sourcemaps in styles causes style-loader to inject
                    // styles using a <link> tag instead of <style> tag. This causes
                    // a FOUC content, which can cause issues with JS that is reading
                    // the DOM for styles (width, height, visibility) on page load.
                    options: {importLoaders: 2, sourceMap: true},
                  },
                  {
                    loader: 'postcss-loader',
                    options: {
                      ident: 'postcss',
                      sourceMap: true,
                      plugins: () => [autoprefixer],
                    },
                  },
                  {loader: 'sass-loader', options: {sourceMap: true}},
                ],
              },
            ],
        },
    },
    prod: {resolve: {alias}},
  },
};

```

Could this be ultimately activated using a flag ?

t-kelly:

Fixes #447 

Allows developers to enable/disable style sourcemaps.

### DIFF
--- a/packages/slate-tools/slate-tools.schema.js
+++ b/packages/slate-tools/slate-tools.schema.js
@@ -88,6 +88,14 @@ module.exports = {
   // https://www.npmjs.com/package/webpack-merge
   'webpack.extend': {},
 
+  // Enabling sourcemaps in styles when using Hot Module Reloading causes
+  // style-loader to inject styles using a <link> tag instead of <style> tag.
+  // This causes a FOUC content, which can cause issues with JS that is reading
+  // the DOM for styles (width, height, visibility) on page load.
+  'webpack.sourceMap.styles': () => {
+    return process.env.NODE_ENV === 'production';
+  },
+
   // Array of PostCSS plugins which is passed to the Webpack PostCSS Loader
   'webpack.postcss.plugins': (config) => [
     autoprefixer,
--- a/packages/slate-tools/tools/webpack/config/parts/__tests__/sass.test.js
+++ b/packages/slate-tools/tools/webpack/config/parts/__tests__/sass.test.js
@@ -18,9 +18,6 @@ test(`passes 'webpack.cssnano.settings' config to cssnano`, () => {
 });
 
 test(`passes 'webpack.postcss.plugins' config to PostCSS Loader`, () => {
-  const oldNodeEnv = process.env.NODE_ENV;
-  process.env.NODE_ENV = 'development';
-
   const SlateConfig = require('@shopify/slate-config');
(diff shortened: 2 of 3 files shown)


adrianocr:
Is this auto-enabled or do we have to add something to the config?

t-kelly:
If you look at the PR, you can see its enabled for production by default. Look at [`slate-tools.schema.js`](https://github.com/Shopify/slate/pull/780/files#diff-adfffc5487594198cfd423cb04028a14R95)

t-kelly:
If you want to always have it enabled, add the following to your `slate.config.js`:
```
'webpack.sourceMap.styles': true,
```

adrianocr:
Awesome! Another question, one I'm not sure fits here or in a separate issue. 

I am using slate with the default starter theme but I removed all the default css and added in Foundation instead. I have a direction in `/src/assets/styles/` named `foundation` and in this foundation dir, I have a gulpfile that compiles the foundation SCSS and outputs it to `/src/assets/styles/app.css` and then in `theme.scss` I import `app.css` which slate uploads and handles as intended. 

My question is once again about sourcemaps. My gulpfile is compiling the CSS with an embedded sourcemap in the the `app.css` file. Will the `'webpack.sourceMap.styles': true,` option have webpack output that sourcemap to the browser or is there something else I need to be doing here to get my sourcemaps outputted to the browser?

t-kelly:
> Will the 'webpack.sourceMap.styles': true, option have webpack output that sourcemap to the browser or is there something else I need to be doing here to get my sourcemaps outputted to the browser?

🤷‍♂️ No clue. You'll have to investigate that one yourself!

Issue #289, opened 2017-09-18
Allow code to be organized in subfolders in `src/snippets` and `src/sections`


t-kelly:
Developers like to keep their repos organized. One way of doing this is through folders and file groupings. This organizational method is possible in folders like `src/styles` and `src/scripts` but not possible in `src/snippets` and `src/sections`.

Let's allow developers to create whatever folder structure they want in the main `src/` folders, and then output those folders in the correct structure for Shopify servers in the `dist/` folder.

For example
```
src/
-- snippets/
---- snippet-1/
------ snippet-1a.liquid
------ snippet-1b.liquid
---- snippet-2/
------ snippet-2a.liquid
------ snippet-2b.liquid
```
would be transformed into the following on build:
```
dist/
-- snippets/
---- snippet-1a.liquid
---- snippet-1b.liquid
---- snippet-2a.liquid
---- snippet-2b.liquid
```

Related https://github.com/Shopify/slate/issues/271

jonathanmoore:
I'd love to see this implemented for development.  The only concern I have would be confusion if someone was using the same file name in more than one directory.

```
src/
-- snippets/
---- blog/
------ tags.liquid
---- collection/
------ tags.liquid
```

Would it make sense if the directory name were prepended to the file name as `[directory].[filename].liquid`?  A quick test verified that `{% include 'test.test' %}` does correctly load `/snippets/test.test.liquid`

The example above would build as:

```
dist/
-- snippets/
------ blog.tags.liquid
------ collection.tags.liquid
```

presto2116:
We are busily redesigning our entire site, and having this feature would be amazing. Our snippets and sections folders are starting to get very large. If we could break it off into folders the development experience will be a lot better. Please  look into adding this feature.

t-kelly:
@presto2116 This feature will be part of Slate v1. We are not adding any more features to Slate v0.

NealEhardt:
>Would it make sense if the directory name were prepended to the file name as `[directory].[filename].liquid`?

When I set up a directory structure, I expect to access it with slashes. So the most sensible way to consume a subfolder snippet would be `{% include 'directory/filename' %}`.

There's a conflict if you perform the proposed dot transform on a directory like this:

    src/
    -- snippets/
    ---- dogs/
    ------ fido.liquid
    ---- dogs.fido.liquid

Maybe as a prerequisite for this issue, Shopify's servers need to start accepting snippets subfolders.

tshamz:
[...]

Issue #336, opened 2017-11-06
Import components for reduced code duplication across projects


t-kelly:
The Shopify themes team has the problem of maintaining 10+ different free themes all with their own unique codebase. Whenever we need to apply an update to our themes, we need to 10+ times. Even worse, each update needs to be customized to each theme because themes development has changed over the past 4 years that they've all been created. 

What if teams could develop and share self-incapsulated components that could be configured and used if a variety of themes. So instead of maintaining 10+ different implementations of a hero slider, we only need to maintain 1 implementation.

Essentially what we need are [Web Components](https://developer.mozilla.org/en-US/docs/Web/Web_Components), but adequate browser support for a fully native implementation is not quite ready. 

One idea to get started is to combine `{% import %}` tags and WebPack to take advantage of sharing snippets (components) through NPM:
```html
<!-- /sections/hello-world.liquid -->
<div id="{{ section.id }}">
  {% import '@ShopifyComponents/hello`, 
     name: section.settings.name, 
     color: section.settings.color 
   %}
</div>
```
```html
<!-- /node_modules/@ShopifyComponents/hello/hello.liquid -->
<h1>Hello, {{ name }}!</h1>

<style>
h1 {
  color: {{ color }}
}
</style>
```
WebPack could parse `{% import %}` tags URL, copy the files referenced to the `/snippets` directory, and change the `{% import %}` URL to reference the newly created snippet. Eg:
````
{% import '@ShopifyComponents/hello` %}
```
would be compiled to
```
{% import 'hello` %}
```

So far this is only a partial implementation of something that kind of resembles Web Components. Would love to hear from the community on any ideas how to approach this problem!

taivu:
this idea sounds and feels like Pattern Lab (https://patternlab.io/).

Drupal uses Twig, which looks similar to Liquid. When I worked with it, i did things like the code examples you shared, like:

```
{% include @scope/thing %}
```

where `@scope` gets mapped to a specific folder. Maybe theres some concepts from Pattern Lab that could influence or inspire some solutions to Shopify Theming.

PR #781, opened 2018-09-26
Allow for subdirectories in `snippets/` and `sections/`

tshamz:
Allow for subdirectories to be used in the `snippets/` and `sections/` directories by updating the `core.js` webpack config in `slate-tools` to use a separate a file-loader rule when dealing with these two directories.

The resolution order for how namespace collisions are handled is done alphabetically by directory name, and from topmost directory to bottom. For example:

- `sections/test.liquid` beats `sections/aaa/test.liquid`
- `sections/zzz/test.liquid` beats `sections/aaa/test.liquid`
- `sections/zzz/zzz/test.liquid` beats `sections/zzz/aaa/test.liquid`
- `sections/zzz/test.liquid` beats `sections/zzz/zzz/test.liquid`

https://github.com/Shopify/slate/issues/289

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/packages/slate-tools/tools/webpack/config/parts/core.js
+++ b/packages/slate-tools/tools/webpack/config/parts/core.js
@@ -104,12 +104,28 @@ module.exports = {
         test: /\.(liquid|json)$/,
         exclude: [
           new RegExp('assets/styles'),
+          new RegExp('sections'),
+          new RegExp('snippets'),
           ...config.get('webpack.commonExcludes'),
         ],
         loader: 'file-loader',
         options: {
           name: '../[path][name].[ext]',
         },
+      },
+      {
+        test: /snippets\/.*\.liquid$/,
+        loader: 'file-loader',
+        options: {
+          name: `../snippets/[name].[ext]`,
+        }
+      },
+      {
+        test: /sections\/.*\.liquid$/,
+        loader: 'file-loader',
+        options: {
+          name: `../sections/[name].[ext]`,
+        }
       },
       {
         test: /assets\/static\//,


tshamz:
not totally sure what's happening here, but if anybody would be willing to give me a little insight/direction, I could start poking around and try fixing this on my own.

dan-gamble:
Should this be able to work like `{% include 'collections/hero' %}` or does the include still need to be written as just `{% include 'hero' %}`?

tshamz:
my preference would be for `{% include 'hero' %}` so that people could write snippets in a style where they initially inherit a core set of snippets and then if they want to use their own, different version of the snippet, instead of creating a snippet with a different name and then updating the reference/include everywhere in their liquid, they just create another one with the same name that overwrites the core one and is automatically used instead, similar to how something like `oh-my-zsh` handles custom plugins/themes.

t-kelly:
My hope is that we can eventually reference snippets and sections with relative paths like:
```
{% include '../snippets/collections/hero.liquid' %}
```
And then on build replace the path with resolved path: 
```
{% include 'hero' %}
```
This would allow us to resolve naming conflicts internally and never let the developer worry about them. For example, if we had these two includes:
```
{% include '../snippets/hero.liquid' %}
{% include '../snippets/collections/hero.liquid' %}
```
they would be resolved to the following on build:
```
{% include 'hero' %}
{% include 'hero2' %}
```
Thoughts?

This PR is moving in that direction, so I'm not opposed to the solution proposed here for now.

tshamz:
tbh, i'm not really that attached to this pull request since everything that it does and that I want it to do can be accomplished from within `slate.config.js`. As I briefly alluded to earlier, what I'm looking to accomplish is set up a base theme that comes with a core set of snippets, let's call them "snippet components, e.g, `button.liquid`, `link.liquid`, `input.liquid`, `hero.liquid`, `list.liquid`, `option-group.liquid`, etc., which are located in `snippets/core/`, and represent the different components used when building a store.

Then, when a developer starts building a new theme, they get all the core snippets and can use them to build out their store, but if they need a component that differs wildly from the version that core snippet component offer, they would made a snippet with the exact same name in a different directory (e.g. `snippets/core/button.liquid` and `snippets/custom/button.liquid`).

Like I mentioned, I can get all of this already with a few modifications to my `slate.config.js`, so it's not a huge bummer to me if this doesn't get merged in. I think the conversation about path resolution and name collisions is interesting and valuable, but not really something this PR set out to accomplish.

t-kelly:
Going to let this one brew for a little since it's not critical and as you mentioned, it can be achieved via the Slate config.

> but if they need a component that differs wildly from the version that core snippet component offer, they would made a snippet with the exact same name in a different directory (e.g. snippets/core/button.liquid and snippets/custom/button.liquid).

What's the advantage of this over just modifying the original? What you're proposing sounds like Wordpress child themes, which existed to facilitate versioning the core/parent theme. 

> I think the conversation about path resolution and name collisions is interesting and valuable, but not really something this PR set out to accomplish.

True. Thanks for clarifying further your goal with this PR.

dan-gamble:
@t-kelly i'm with you in that i definitely feel you should be using some sort of patch to include things as opposed to magic resolving.

Only think i'd potentially change is instead of

```liquid
{% include 'hero' %}
{% include 'hero2' %}
```

It would be something like:

```liquid
{% include 'hero' %}
{% include 'collectionshero' %}
```

I see "built" liquid files as slightly different to built CSS / JS. They can still somewhat be used to read and don't get obscured as much as JS / CSS so keeping it semi readable would be good imo

tshamz:
> What's the advantage of this over just modifying the original? What you're proposing sounds like Wordpress child themes, which existed to facilitate versioning the core/parent theme. Shopify themes don't really have anything similar to this, nor does Slate?

My main motivation here was to give devs in our organization a set of building blocks written in a way that prevents from making errors and comes prebuilt with functionality already baked in. The moment people start altering the core snippet library, bad things are liable to happen. Versioning the core snippets with something like a submodule was something I was also interested in as well. However, after reading you inputs and thinking about it, I might pivot on this a bit.

Anyways, as has been mentioned a few times before, I can get what I want without this PR so feel to do whatever you'd like with it. Merge it, close it, ridicule it, worship it, whatever, I won't lose any sleep over it.

t-kelly:
> Versioning the core snippets with something like a submodule was something I was also interested in as well.

Sounds a little like https://github.com/Shopify/slate/issues/336

Issue #804, opened 2018-10-08
Slate analytics prompting email even if already exists Continuous Integration.


diegopego86:
I'm using Team City for contiues integration. Runing yarn deploy --env=qa --skipPrompts for example, slate analytics still prompts:

```
To continue, please enter your email address:
```

I noticed a few issues here:

1- In slate-theme we are running 2 tasks on deploy, but yarn:
slate-tools build && slate-tools deploy
As in this link https://stackoverflow.com/questions/50835221/pass-command-line-argument-to-child-script-in-yarn it seems that the arguments passed in yarn are only added to the last part of the script, so what is happening in my case is :

```
slate-tools build && slate-tools deploy --env=qa --skipPrompts
```

So the environment and skipPrompts arguments are only passed to deploy and not to the build command.

2- On slate-analytics/index.js line 24:

```
let email = getUserEmail();
```
This value is never read from the env file, first because of the issue 1, but even fixing that error won't help. For some reason the env is not loaded correctly and the function always return undefined.

3 - SLATE_USER_EMAIL, it seems that we can use this environment variable but there is no information about it.

### Replication steps

1) First you need to remove your ~/.slaterc file as this contains the email you've set up when you first created your project

```
rm -r ~/.slaterc
```

2) Add SLATE_USER_EMAIL=  into you .env.{environment} file assigning a valid email to it

.env=qa
```
SLATE_USER_EMAIL=mytest@email.com
```

3) Run 
```
yarn deploy --env=dev-qa --skipPrompts
```

You'll now should be prompted for an email:

![screen shot 2018-10-08 at 12 05 14 pm](https://user-images.githubusercontent.com/1388364/46589166-6e866d00-caf2-11e8-8c14-d3865580d126.png)

This shouldn't happen as the email has already been set in .env=qa

I'm creating some pull request with the updates I made to try to fix this issue.

marcnewport:
+1 for buddy

t-kelly:
Good catch! 💯

PR #805, opened 2018-10-08
=Add custom deploy script on slate-tools to run pre-run a build scrip…

diegopego86:
…t. Update slate-analitycs to load the environment passed in the arguments. Update readme file to add SLATE_USER_EMAIL variable

Trying to fix issue #804 

### Checklist
For contributors:
- [ ] I have [updated the readme](https://github.com/Shopify/slate/tree/master/packages/slate-env) to add SLATE_USER_EMAIL variable.

For maintainers:
- [ ] I have added a custom 'deploy' task in packages/slate-tools/cli/index.js to run a build before the deploy. This is to be able to remove the multitasks in the [package.json](https://github.com/Shopify/starter-theme/blob/master/package.json) on the Starter Theme. This is because yarn can only append the arguments to the last script, so in the case of deploy, the first script before the &&:
```
slate-tools build
```

Will not load the --env=qa data.

- [] I have updated the index.js file in slate-analytics to load correctly the slateEnv, In my case, the getUserEmail() was never loading the email in the environmnet file see #804 

-[] For this fix it is also important to update https://github.com/Shopify/starter-theme, check my pull request https://github.com/Shopify/starter-theme/pull/113

### DIFF
--- a/packages/slate-analytics/index.js
+++ b/packages/slate-analytics/index.js
@@ -3,10 +3,10 @@ const clearConsole = require('react-dev-utils/clearConsole');
 const rc = require('@shopify/slate-rc');
 const slateEnv = require('@shopify/slate-env');
 const axios = require('axios');
+const argv = require('yargs').argv;
 const prompt = require('./prompt');
 const {validateEmail} = require('./utils');
 const packageJson = require('./package.json');
-const argv = require('yargs').argv;
 
 async function init() {
   let config = rc.get() || rc.generate();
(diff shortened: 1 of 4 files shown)


t-kelly:
Great issue and PR @diegopego86. Both really helped me understand the problem!

I created a followup PR with an adjusted approach. Please see the note at the bottom of the description on how to solve the problem of SLATE_USER_EMAIL not being set for `yarn deploy`

Issue #702, opened 2018-08-13
SLATE_IGNORE_FILES does not work


zapsilere:
### Problem

SLATE_IGNORE_FILES does not ignore my files

### Replication steps

add this to your .env
SLATE_IGNORE_FILES=/config/settings_data.json:/assets/static/icons.svg

### More Information
I want to ignore my icons.svg as whenever I deploy my work the icons.svg wont work properly so I have to upload it to the store directly whenever I deploy

am I writing the proper syntax?

jonathanmoore:
@arthurgouveia for your second file are you also trying to ignore a file in the assets folder?

This did work for me successfully:
SLATE_IGNORE_FILES=/config/settings_data.json:/layout/theme.liquid

zapsilere:
@jonathanmoore have you tried on assets?

t-kelly:
This may be a regression caused by https://github.com/Shopify/slate/pull/647. Will investigate!

PR #814, opened 2018-10-10
Clarify how the the SLATE_IGNORE_FILES env variable works

t-kelly:

Followup to #702 

cc @huguestennier can we make sure any other docs we have on this are updated too?


huguestennier:
I think we need to clarify in the docs that the ignored files will still be processed and put in the `dist` folders. They only get ignored on `deploy`. Maybe rename this to something like `SLATE_IGNORE_FILES_ON_DEPLOY` ?

Gonna update the wiki right away because it says to look for the path in `src` folder.

PR #821, opened 2018-10-13
Prevent trailing slashes at the end of store URLs

### LINKED ISSUE #772: Unhelpful ENOTFOUND error when passing invalid SLATE_STORE value
### Problem

Followup to https://github.com/Shopify/slate/issues/652 and https://github.com/Shopify/slate/issues/629 

Looks like if you pass a URL with a trailing `/`, users get an unhelpful error. This error could be more informative!

maximevaillancourt:
Fix #772 by surfacing an error that explicitly points out that store URLs should not include a trailing slash.

### Checklist
For contributors:
- [X] I have made sure that the documentation [mentions not including the trailing slash](https://github.com/Shopify/slate/wiki/3.-Connect-to-your-store#setting-slate_store).

For maintainers:

### DIFF
--- a/packages/slate-env/__tests__/index.test.js
+++ b/packages/slate-env/__tests__/index.test.js
@@ -328,6 +328,20 @@ describe('Slate Env', () => {
         );
       });
 
+      test('the store URL environment variable does not end with a trailing slash', () => {
+        ['shop1.myshopify.com', 'shop1.myshopify.com/'].forEach((value) => {
+          setVars(
+            Object.assign({}, TEST_ENV, {
+              [config.get('env.keys.store')]: value,
+            }),
+          );
+          const result = slateEnv.validate();
+          expect(result.errors).toHaveLength(
+            value === 'shop1.myshopify.com/' ? 1 : 0,
+          );
+        });
+      });
+
       test('the store API password environment variable is empty', () => {
         setVars(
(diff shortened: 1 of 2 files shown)


dan-gamble:
Is there any benefit to automatically fixing this instead of requiring the user to remove the slash themselves?

maximevaillancourt:
> Is there any benefit to automatically fixing this instead of requiring the user to remove the slash themselves?

Sounds like a good idea. I see two options:
1. Slate raises an error when a trailing slash is detected, and asks the user to remove it themselves (current implementation)
2. Slate removes the trailing slash automatically, _and communicates the automatic fix to the user_

The latter part of option 2 is what's important. As a developer, I'd want to know what is wrong with the config so that I don't repeat the mistake later on. If Slate lets me know about the issue and the suggested fix, _and_ also fixes the issue automatically, I'm all for it.

@t-kelly, what's your take on this?

t-kelly:
> As a developer, I'd want to know what is wrong with the config so that I don't repeat the mistake later on. If Slate lets me know about the issue and the suggested fix, and also fixes the issue automatically, I'm all for it.

I like the sound of this. Throw a warning, not an error.

t-kelly:
Going to merge this for now since it's better than what we have now (aka nothing). Would love to see a followup PR @maximevaillancourt!

PR #838, opened 2018-10-16
Add SSL check script to a convenient yarn command

tonyd256:

The ssl-check script that @callaginn wrote is super useful and fixes the SSL issues I was having as well as others have been having. The current docs say to add this script inside your profile or where your other local scripts reside; however, it would be way more convenient to provide it with the library while it's still needed.

This will require a docs change for sure, but hopefully it will be more deletion than additions.

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "bootstrap": "yarn install && node_modules/.bin/lerna bootstrap",
     "publish": "node_modules/.bin/lerna publish --exact",
     "changelog": "node_modules/.bin/lerna-changelog",
-    "clean": "node_modules/.bin/lerna clean"
+    "clean": "node_modules/.bin/lerna clean",
+    "dev": "bin/ssl-check && yarn start"
   },
   "devDependencies": {
     "babel-jest": "^22.0.4",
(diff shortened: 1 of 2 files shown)


tonyd256:
Just signed the CLA.

Also might be useful to get this in there: `brew install mkcert && mkcert -install`

package.json
```diff
@@ -23,7 +23,8 @@
     "bootstrap": "yarn install && node_modules/.bin/lerna bootstrap",
     "publish": "node_modules/.bin/lerna publish --exact",
     "changelog": "node_modules/.bin/lerna-changelog",
-    "clean": "node_modules/.bin/lerna clean"
+    "clean": "node_modules/.bin/lerna clean",
+    "dev": "bin/ssl-check && yarn start"
```
t-kelly: I don't think you've tested this...

tonyd256: I added it locally, sorry just copied over and I think you're right that it's not going to work since I added it to my package.json. Really this needs to be in the generated package.json or linked else where. Sorry for the non-working PR. I will work on putting it in the right place.

t-kelly:
Closing due to inactivity

Issue #863, opened 2018-11-01
Incorrect environment displayed in themekit's logs when using non-dev .env files


HauntedSmores:
### Problem

Themekit shows it's using `development` environment regardless of `.env.***` file.

While the new Slate warnings are great and remind you which theme you're on, it does not inform you of the correct environment. The reason for this is that when Slate constructs the `config.yml` file in the dist folder, it simply puts whatever was in your `.env` file as the `development` entry.

This gave me great pause when I was deploying using `.env.staging` and seeing:

```
[development]: 50 / 100 [--------------->-------------] 50%
```

### Replication steps

Simply run `npm run deploy -- --env=anything` and it will show the above console log (`[development]`)

### More Information

Build process should take `****` from `.env.****` and use that as the key name in the constructed `config.yml` - if using just `.env`, then leave env name as `development`.

t-kelly:
Fixed in #866

PR #866, opened 2018-11-02
Add environment to deploy message

huguestennier:

Partially fixing #863.

### Problem

When Themekit is deploying on Shopify, it will show progress like this : 

`[environment]: 10 / 30 [=====================>--------------]  33 %`

which works great when you are using the default `config.yml` files. 

Since we are using `.env` files, Themekit doesn't understand that there's multiple environment. So we always show `[development]` even if we are using another environment.

If we try to add the `--env` flag to our [_generateConfigFlags()](https://github.com/Shopify/slate/blob/5c68081db54cd94c79775d02bc10069676e2fde6/packages/slate-sync/index.js#L44), Themekit will complain that there's no 'staging' environment, for example : `staging does not exist in this environments list`.

### Solution

I suggest adding the environment in the `↑  Uploading to Shopify...` message above as a workaround.

The developer will now see : 

```
↑  Uploading to Shopify on staging environment...
[development]: 10 / 30 [=====================>--------------]  33 %
```

I don't think it's perfect, but it works for now.

### DIFF
--- a/packages/slate-sync/index.js
+++ b/packages/slate-sync/index.js
@@ -82,7 +82,13 @@ async function deploy(cmd = '', files = []) {
 
   deploying = true;
 
-  console.log(chalk.magenta(`\n${figures.arrowUp}  Uploading to Shopify...\n`));
+  console.log(
+    chalk.magenta(
+      `\n${
+        figures.arrowUp
+      }  Uploading to Shopify on ${slateEnv.getEnvNameValue()} environment...\n`,
+    ),
+  );
 
   try {
     await promiseThemekitConfig();


huguestennier:
> promiseThemekitConfig() uses themekit to generate a config.yml file. We could generate it ourselves so that we can specify the environment name?

🤦‍♂️ I swear I tried that and it wasn't working. Fixed.

huguestennier:
`.env` 

screen shot 2018-11-12 at 10 40 49 am

`.env.staging`

screen shot 2018-11-12 at 10 41 08 am

PR #947, opened 2019-01-15
Update Commands.md

george-dotdev:

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/docusaurus/v1/versioned_docs/version-1.0.0-beta.14/Commands.md
+++ b/docusaurus/v1/versioned_docs/version-1.0.0-beta.14/Commands.md
@@ -18,7 +18,7 @@ The following is a list of all available API commands for Slate Tools:
 Builds a production-ready version of the theme by compiling the files into the `dist` folder.
 
 ```bash
-slate-tools build
+yarn build
 ```
 
 Sets `process.env` to `’production’`, which can be used in Slate config to return production only values, e.g.:


t-kelly:
I see what you're trying clarify here, but I don't think this is right way.

`yarn build` is the bash command you use to call the `build` script which is specified in any project's package.json file. The contents of the package.json `build` script can be anything and ultimately is specified by the theme, not Slate. For example, if someone wanted to do something else to their theme before running `slate-tools build`, they could modify the `build` script to run another script before, e.g. `build: 'node my-custom-script.js && slate-tools build`.

`slate-tools build` is the bash command made available in the `node_modules/.bin` folder of your project when `slate-tools` is installed as a dependency. It calls Slate's build script and does what we talk about in this doc. It can be called by referencing it directly in a `package.json` file, like you see in Starter Theme, or via `npx slate-tools build` when called in the same directory as your theme.

What I think this dock might need is something in the introduction explaining how these commands are typically used, aka 'within a package.json file of a Slate theme', and maybe an example and link to Shopify Starter Theme's package.json for an example?

jonathanmoore:
Here is my opinion on how to structure it... Currently, the Commands page inside Getting Started focuses on the commands of the behind-the-scenes `slate-tools` package.  In the Getting Started section I would make the assumption that the commands would cover the default commands in the package.json file and how to add flags.

- `yarn start`
- `yarn watch`
- `yarn build`
- `yarn deploy`
- `yarn zip`
- `yarn lint`
- `yarn lint-json`
- `yarn format`

Having documentation for `slate-tools` commands is still important for developers customizing their Slate implementation.  I would expect to find those specific details in the Reference section, along with basic examples pointing to the `package.json` file.

PR #967, opened 2019-01-31
Resolve Windows Slashes for styles and script tags snippets.

### LINKED ISSUE #966: Generated script-tags for customer templates not checking Windows Slashes
### Problem

See #806 - Resolved issue does not work for Windows under `slate-tools/tools/webpack/script-tags.html`, since;
`<% if (htmlWebpackPlugin.options.liquidTemplates[chunk].includes('customers/')) { %>`
will be `false` on windows due to directory slashes being `\` rather than `/`

### Replication steps

1. Pull Slate starter theme on a Windows based PC
2. Execute `yarn build`
3. Inspect `dist/snippets/script-tags.liquid` to see that the template if statement will be as: `{%- if template == 'login' -%}`

### More Information

Easily fixed by replacing `/` with `\\` on Windows, ideally needs to facilitate both environments.

YourWishes:

Resolves #966 

Added extra check for Windows Slashes in the script and style tags snippet template.

### DIFF
--- a/packages/slate-tools/tools/webpack/script-tags.html
+++ b/packages/slate-tools/tools/webpack/script-tags.html
@@ -12,13 +12,15 @@
     {%- else -%}
       <link rel="prefetch" href="<%= src %>" as="script">
     {%- endif -%}
+    
   <% } else if (chunk.split('@').length > 1) { %>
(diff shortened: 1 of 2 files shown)


t-kelly:
Thanks for the fix!

PR #954, opened 2019-01-18
Now sections can have separate .json file for schema.

MikhailRoot:
## Now sections can have separate .json file for schema.

If  we have `section_name.schema.json` file in same directory as `section_name.liquid` file  it will inject it as a `{% schema %}` for this section while placing it  in `/dist/sections` directory.

It checks if `section_name.liquid` has inline definition and does nothing if it has it inline.
If `section_name.liquid` don't have neither `section_name.schema.json` neither inline `{% schema %}` declaration in it, it outputs a warning.

The win of this approach is simplifying json editing leading to less errors in schema declaration. Also you have good syntax highlighting in editor of choice for .json files for sure.

https://github.com/Shopify/slate/issues/749
https://github.com/Shopify/slate/issues/856

### Checklist
For contributors:
- [ ] I have [updated the docs](https://github.com/Shopify/slate/blob/master/CONTRIBUTING.md#documentation) to reflect these changes, if applicable.

For maintainers:
- [ ] I have :tophat:'d these changes.

PR #982, opened 2019-02-19
[WIP] Slate Sections Plugin

harshal317:

In order to support a sections format that allows separating the schema from the liquid, and allowing a structure that makes it easy to maintain translations for many languages, @t-kelly  and I came to a conclusion it may be a good idea to move away from CopyWebpackPlugin. Especially keeping in mind the potential for additional features to this plugin. 

This plugin aims to support three basic structures
1) Having a sections folder consisting of simply liquid files, which will simply be copied to the dist sections path

2) Having a sections folder consisting of folders for each section following the naming convention /sections/example-section-name/ having a mandatory 'template.liquid' file, and an optional 'schema.json' file. If there is a schema file in the directory it will get appended to the liquid file adding the schema tags, and will be outputted to the dist sections path taking the name of the directory (example-section-name.liquid)

3) Similiar to option 2, however it extends the functionality by adding a locales folder which can have multiple json files for each language. The translations will automatically be added to the json object that gets appended to the liquid making it easier to maintain multiple languages and keeping a cleaner codebase

### To Do
- [ ] Adding more robust tests,
- [ ] Adding a README
- [ ] Taking in and improving the plugin based on feedback
- [ ] Adding error handling

### Checklist
For contributors:

For maintainers:


harshal317:
The dirent class was added in Node 10, so might have to change that logic slightly to support older versions.

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+          path.resolve(compilation.options.context, 'sections'),
+          compilation,
+          dirents,
+        );
+      }
+    });
+  }
+};
+
+async function getSchema(liquidDirent, directoryPath, compilation, dirents) {
```
harshal317: I feel like the logic in this could definitely be simplified, maybe some fresh eyes will help with that

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+const path = require('path');
+const {ConcatSource, RawSource} = require('webpack-sources');
+const _ = require('lodash');
+
+module.exports = class localesPlugin {
+  constructor(options) {
+    this.options = options;
+  }
+  apply(compiler) {
+    compiler.hooks.compilation.tap('Slate Locales Plugin', this.addLocales);
```
t-kelly: Rename.

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  async addLocales(compilation) {
+    const dirents = await fs.readdir(
```
t-kelly: `dirents`?

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+      path.resolve(compilation.options.context, 'sections'),
```
t-kelly: You should be pulling this path from `slate-config`

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+      {withFileTypes: true},
+    );
+    // eslint-disable-next-line prefer-arrow-callback
+    dirents.forEach((dirent) => {
+      if (dirent.isDirectory()) {
+        traverseDirectory(
+          dirent,
+          path.resolve(compilation.options.context, 'sections'),
```
t-kelly: Pull from `slate-config`

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+          compilation,
+          dirents,
+        );
+      }
+    });
+  }
+};
+
+async function getSchema(liquidDirent, directoryPath, compilation, dirents) {
+  if (liquidDirent.name === 'template.liquid') {
```
t-kelly: Hard coded strings should at least be set as constants up top.

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+    const outputKey = `../sections${
```
t-kelly: Does this have to be a relative path? This will break if we change the `output` of our webpack config. Seems fragile.

packages/slate-locales-plugin/index.js
t-kelly: This method is doing a lot more than just `getSchema` -- can you work on scoping the contents into logical methods to improve overall readability? The nesting is really hard to read.

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+      directoryPath.split('/sections')[1]
+    }.liquid`;
+
+    if (
+      dirents.find((obj) => obj.name === 'locales' && obj.isDirectory()) &&
+      dirents.find((obj) => obj.name === 'schema.json' && obj.isFile())
+    ) {
+      compilation.assets[outputKey] = new ConcatSource(
```
t-kelly: This is really gnarly to read -- can we clean it up?

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+        new RawSource(
+          fs.readFileSync(path.resolve(directoryPath, 'template.liquid')),
```
t-kelly: Can we use the async version for improved performance?

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+        ),
+        new RawSource(
+          `{% schema %}\n ${JSON.stringify(
```
t-kelly: You've got a 3 nested methods here. This could be simplified.

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+              combineLocales(path.resolve(directoryPath, 'locales')),
+              path.resolve(directoryPath, 'schema.json'),
+            ),
+            null,
+            2,
+          )}{% endschema %}`,
+        ),
+      );
+    } else if (
+      dirents.find((obj) => obj.name === 'schema.json' && obj.isFile())
```
t-kelly: These conditions are really vague at indicating what transformations you're about to do.

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+    ) {
+      compilation.assets[outputKey] = new ConcatSource(
```
t-kelly: This looks very similar to above. Reduce duplication, reuse code.

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+    );
+    const outputKey = `../sections${
+      directoryPath.split('/sections')[1] === ''
+        ? `/${liquidDirent.name}`
+        : `${directoryPath.split('/sections')[1]}/${liquidDirent.name}`
+    }`;
+    if (
+      liquidContent.match(/{%\s*schema\s*%}([\s\S]+){%\s*endschema\s*%}/gim) ||
+      !dirents.find(
+        (obj) => obj.name === liquidDirent.name.replace('.liquid', '.json'),
```
t-kelly: What's the reason for looking for a json file of the same name?

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+}
+
+async function traverseDirectory(dirent, directoryPath, compilation) {
+  const dirents = await fs.readdir(path.resolve(directoryPath, dirent.name), {
+    withFileTypes: true,
+  });
+  dirents.forEach((subdirent) => {
+    if (subdirent.isDirectory() && subdirent.name !== 'locales') {
```
t-kelly: I think we should only allow one layer of nesting, no? What's the point of supporting more?

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+        obj[i] = getLocalizedValues(obj[i].t, obj, localizedSchema);
+      } else if (typeof obj[i] === 'object') {
+        traverse(obj[i]);
+      }
+    }
+  }
+  // eslint-disable-next-line
+  let mainSchema = JSON.parse(fs.readFileSync(mainSchemaPath, 'utf-8'));
+  traverse(mainSchema);
+  return mainSchema;
```
t-kelly: Would be cleaner if `traverse()` returned `obj` so you could just have `return traverse()`.

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+}
+
+function combineLocales(localesPath) {
+  return fs
+    .readdirSync(localesPath)
```
t-kelly: Async

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
...
+    .filter((fileName) => fileName.endsWith('.json'))
+    .reduce((accumulator, file) => {
+      const localeCode = path
+        .basename(file)
+        .split('.')
+        .shift();
+      const fileContents = JSON.parse(
+        fs.readFileSync(path.resolve(localesPath, file), 'utf-8'),
```
t-kelly: Async

packages/slate-locales-plugin/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "slate-locales-plugin",
```
t-kelly: `slate-sections-plugin`

packages/slate-locales-plugin/playground.js
```diff
@@ -0,0 +1,9 @@
+const compiler = require('./__tests__/helpers/compiler');
```
t-kelly: Remove.

packages/slate-tools/package.json
```diff
@@ -68,6 +68,7 @@
     "raw-loader": "^0.5.1",
     "react-dev-utils": "0.5.2",
     "sass-loader": "6.0.7",
+    "slate-locales-plugin": "file:../slate-locales-plugin",
```
t-kelly: Set to `1.0.0` and run `yarn bootstrap` from the root directory to link local packages.

packages/slate-tools/tools/webpack/config/prod.js
```diff
@@ -6,7 +6,7 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
 const SlateConfig = require('@shopify/slate-config');
 const SlateTagPlugin = require('@shopify/slate-tag-webpack-plugin');
-
+const SlateLocalesPlugin = require('slate-locales-plugin');
```
t-kelly: Keep row space below

packages/slate-tools/tools/webpack/config/prod.js
```diff
@@ -85,6 +85,7 @@ module.exports = merge([
       new HtmlWebpackIncludeLiquidStylesPlugin(),
 
       new SlateTagPlugin(packageJson.version),
+      new SlateLocalesPlugin({}),
```
t-kelly: Why the empty object?

packages/slate-locales-plugin/__tests__/compiler-tests.js
```diff
@@ -0,0 +1,54 @@
+const path = require('path');
+const compiler = require('./helpers/compiler');
+
+jest.unmock('fs-extra');
+
+test('starter theme structure', async () => {
```
t-kelly: More descriptive test names please. This tells me very little.

packages/slate-locales-plugin/index.js
```diff
@@ -0,0 +1,183 @@
+const fs = require('fs-extra');
+const path = require('path');
+const {ConcatSource, RawSource} = require('webpack-sources');
+const _ = require('lodash');
+
+module.exports = class localesPlugin {
+  constructor(options) {
+    this.options = options;
```
t-kelly: Unused?

packages/slate-locales-plugin/index.js
harshal317: The functionality https://github.com/Shopify/slate/pull/954 this PR wanted

packages/slate-locales-plugin/index.js
harshal317: Wasn't initially sure if people nested but if you think we don't need further nesting then that should make it simpler

packages/slate-locales-plugin/index.js
harshal317: Simplified the logic, it was pretty nasty

t-kelly:
From @davidwarrington:

> Would it be possible to add `.js` file support to this? Being able to import a schema exported from a JS file would open up a lot of possibilities. For example; querying an API to retrieve a list of options, or simply being able to build multiple similar options with a for loop. Most stores probably don't require anything like this however I have worked on a few stores where this would be beneficial.

> The way I see this working is by using `const jsonSchema =require(jsonSchemaFilePath)` in line 22 of `inject-schemas-to-sections.js`. Since Node caches modules, they would need to be flushed out of the cache when updating, in order to be re-imported, but that's not a particularly difficult job.

harshal317:
> From @davidwarrington:
> 
> > Would it be possible to add `.js` file support to this? Being able to import a schema exported from a JS file would open up a lot of possibilities. For example; querying an API to retrieve a list of options, or simply being able to build multiple similar options with a for loop. Most stores probably don't require anything like this however I have worked on a few stores where this would be beneficial.
> 
> > The way I see this working is by using `const jsonSchema =require(jsonSchemaFilePath)` in line 22 of `inject-schemas-to-sections.js`. Since Node caches modules, they would need to be flushed out of the cache when updating, in order to be re-imported, but that's not a particularly difficult job.

This is definitely something that can be done, however it is probably worth discussing whether we want to do it in this current PR or as a feature moving forward.

packages/slate-tools/tools/webpack/config/prod.js
```diff
@@ -6,6 +6,7 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const UglifyJSPlugin = require('uglifyjs-webpack-plugin');
 const SlateConfig = require('@shopify/slate-config');
 const SlateTagPlugin = require('@shopify/slate-tag-webpack-plugin');
+const SlateLocalesPlugin = require('slate-locales-plugin');
```
t-kelly: You'll want to add this to `core.js` since we want to do the transformation in both dev and prod builds.

packages/slate-sections-plugin/index.js
```diff
@@ -0,0 +1,204 @@
...
+          .split('.')
+          .shift();
+        const fileContents = JSON.parse(
+          await fs.readFile(path.resolve(localesPath, file), 'utf-8'),
+        );
+        accumulator[localeCode] = fileContents;
+      }),
+    );
+
+    return accumulator;
```
t-kelly: I think you could alternatively write:
```
return await jsonFiles.reduce(async (promise, file) => {
  const accumulator = await promise;
  ...
  return accumulator[localCode] = await fs.readJson(path.resolve(localesPath, file));
}, Promise.resolve({}));
```

Let me know if it works? 😅

packages/slate-sections-plugin/index.js
```diff
@@ -0,0 +1,204 @@
+const fs = require('fs-extra');
+const path = require('path');
+const {ConcatSource, RawSource} = require('webpack-sources');
+const _ = require('lodash');
+
+// If section liquid file is the GENERIC_TEMPLATE_NAME the output liquid will take the form of directoryName.liquid
+const GENERIC_TEMPLATE_NAME = 'template.liquid';
```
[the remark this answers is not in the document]
harshal317: Reasonable option to give users 👍

packages/slate-sections-plugin/index.js
```diff
@@ -0,0 +1,204 @@
...
+      for (const i in obj) {
+        if (typeof obj[i].t === 'string') {
+          obj[i] = this._getLocalizedValues(obj[i].t, obj, localizedSchema);
+        } else if (typeof obj[i] === 'object') {
+          traverse(obj[i]);
+        }
+      }
+      return JSON.stringify(obj);
+    };
+    const mainSchema = JSON.parse(await fs.readFile(mainSchemaPath, 'utf-8'));
```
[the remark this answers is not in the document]
harshal317: Oops TIL

packages/slate-sections-plugin/__tests__/index.test.js
```diff
@@ -0,0 +1,68 @@
+const path = require('path');
+const compiler = require('./helpers/compiler');
+
+jest.unmock('fs-extra');
```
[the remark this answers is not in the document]
harshal317: Also annoying that fs doesn't get mocked by default but fse does

packages/slate-sections-plugin/index.js
harshal317: Worked 👍

packages/slate-sections-plugin/index.js
```diff
@@ -0,0 +1,220 @@
...
+    const relativePathFromSections = path.relative(
+      this.options.from,
+      liquidSourcePath,
+    );
+
+    const fileName = this._getOutputFileName(relativePathFromSections);
+
+    // The relative path from the output set in webpack, to the specified output for sections in slate config
+    const relativeOutputPath = path.relative(
+      compilation.compiler.outputPath,
```
t-kelly: You could pass just this value instead of the whole compilation object

packages/slate-sections-plugin/index.js
```diff
@@ -0,0 +1,220 @@
...
+    await Promise.all(
+      files.map(async (file) => {
+        const fileStat = await fs.stat(path.resolve(this.options.from, file));
+        if (fileStat.isDirectory()) {
+          const pathToLiquidFile = path.resolve(
+            this.options.from,
+            file,
+            this.options.genericTemplateName,
+          );
+          const outputKey = this._getOutputKey(pathToLiquidFile, compilation);
```
t-kelly: You could declare up top `const compilationOutput = compilation.compiler.outputPath` and pass it here

packages/slate-sections-plugin/index.js
```diff
@@ -0,0 +1,220 @@
...
+   * @param {*} localesPath Absolute path to the locales folder /sections/section-name/locales/
+   * @returns
+   */
+  async _combineLocales(localesPath) {
+    const localesFiles = await fs.readdir(localesPath);
+    const jsonFiles = localesFiles.filter((fileName) =>
+      fileName.endsWith('.json'),
+    );
+
+    const combinedLocales = await jsonFiles.reduce(async (promise, file) => {
```
t-kelly: you can just return this, no need for the constant assignment

packages/slate-tools/tools/webpack/config/parts/core.js
```diff
@@ -104,5 +101,10 @@ module.exports = {
         to: config.get('paths.theme.dist.templates'),
       },
     ]),
+    new SlateLocalesPlugin({
+      from: config.get('paths.theme.src.sections'),
+      to: config.get('paths.theme.dist.sections'),
+      genericTemplateName: 'template.liquid',
```
t-kelly: Maybe we can have this set as the default value instead of needing to declare it? It can remain an option if users want to specify another name.

packages/slate-sections-plugin/index.js
```diff
@@ -0,0 +1,239 @@
...
+          );
+          const outputKey = this._getOutputKey(
+            pathToLiquidFile,
+            compilationOutput,
+          );
+          compilation.assets[
+            outputKey
+          ] = await this._getWebpackSourceForDirectory(
+            path.resolve(this.options.from, file),
+            compilation,
```
t-kelly: This isn't used. Remove.

packages/slate-sections-plugin/__tests__/index.test.js
```diff
@@ -0,0 +1,111 @@
...
+  expect(
+    JSON.parse(
+      stats.compilation.assets[expectedAssetOutputKey].children[1]._value
+        .replace('{% schema %}', '')
+        .replace('{% endschema %}', ''),
+    ),
+  ).toMatchSnapshot();
+});
+
+test('throws type error if from missing from options', () => {
```
t-kelly:
```suggestion
test('throws type error if `from` key is missing from options', () => {
```

packages/slate-sections-plugin/__tests__/index.test.js
```diff
@@ -0,0 +1,111 @@
...
+  const options = {
+    to: path.resolve(__dirname, '../dist/sections'),
+  };
+  const errorToBeThrown = new TypeError('Missing or Invalid From Option');
+  expect(() => {
+    new SlateSectionsPlugin(options);
+  }).toThrowError(errorToBeThrown);
+});
+
+test('throws type error if from option is not string', () => {
```
t-kelly:
```suggestion
test('throws type error if `from` option is not string', () => {
```

packages/slate-sections-plugin/__tests__/index.test.js
```diff
@@ -0,0 +1,111 @@
...
+    from: 10,
+    to: path.resolve(__dirname, '../dist/sections'),
+  };
+  const errorToBeThrown = new TypeError('Missing or Invalid From Option');
+  expect(() => {
+    new SlateSectionsPlugin(options);
+  }).toThrowError(errorToBeThrown);
+});
+
+test('throws type error if to missing from options', () => {
```
t-kelly:
```suggestion
test('throws type error if `to` key is missing from options', () => {
```

packages/slate-sections-plugin/__tests__/index.test.js
```diff
@@ -0,0 +1,111 @@
...
+  const options = {
+    from: path.resolve(__dirname, '../src/sections'),
+  };
+  const errorToBeThrown = new TypeError('Missing or Invalid To Option');
+  expect(() => {
+    new SlateSectionsPlugin(options);
+  }).toThrowError(errorToBeThrown);
+});
+
+test('throws type error if to option is not string', () => {
```
t-kelly:
```suggestion
test('throws type error if `to` option is not string', () => {
```

packages/slate-sections-plugin/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "slate-sections-plugin",
+  "version": "1.0.0",
```
t-kelly: Should be `1.0.0-beta.15` so when we release it's part of the beta.16 release

packages/slate-tools/package.json
```diff
@@ -68,6 +69,7 @@
     "raw-loader": "^0.5.1",
     "react-dev-utils": "0.5.2",
     "sass-loader": "6.0.7",
+    "slate-sections-plugin": "1.0.0",
```
t-kelly: `"@shopify/slate-sections-plugin": "1.0.0-beta.15"`

packages/slate-sections-plugin/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "slate-sections-plugin",
```
t-kelly: `@shopify/slate-sections-plugin`

packages/slate-tools/tools/webpack/config/parts/core.js
```diff
@@ -4,6 +4,7 @@ const CopyWebpackPlugin = require('copy-webpack-plugin');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const SlateConfig = require('@shopify/slate-config');
 const config = new SlateConfig(require('../../../../slate-tools.schema'));
+const SlateLocalesPlugin = require('slate-locales-plugin');
```
t-kelly: Update this.

packages/slate-tools/tools/webpack/config/parts/core.js
```diff
@@ -104,5 +101,9 @@ module.exports = {
         to: config.get('paths.theme.dist.templates'),
       },
     ]),
+    new SlateLocalesPlugin({
```
t-kelly: Add blank line above this. Update name.

packages/slate-sections-plugin/README.md
```diff
@@ -0,0 +1,166 @@
+# @shopify/slate-sections-plugin
+
+The plugin handles the taking your sections folder, and outputting files that will work in production. There are 2 supported sections folder structures.
```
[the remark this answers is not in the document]
harshal317: Much better worded than myself

dan-gamble:
This is turning in to an awesome PR 👏

Can't wait to start using this on projects

dan-gamble:
Since folders are looking to be introduced here would it be much extra effort to allow an additional layer of folders in the `sections` folder? For example:

```
./sections
├── homepage/
   ├── featured-collections/
      ├── schema.json
      ├── template.liquid
   ├── hero/
      ├── schema.json
      ├── template.liquid
   ├── ...
├── product/
   ├── featured-articles/
      ├── schema.json
      ├── template.liquid
   ├── variants/
      ├── schema.json
      ├── template.liquid
   ├── ...
```

harshal317:
> This looks great! @t-kelly made my life easier with his early reviews (thanks). I still need to 🎩but I'll do that first thing tomorrow morning.
> 
> Side note: did you and @t-kelly discuss the possibility of migrating the locale logic over so we can use it for merchant-facing translations for `config/settings_schema.json`, not just for sections? This would enable a similar folder structure to how we're structuring section folders and enable translations to be referencing with the `t` key:
> 
> ```shell
> ./config
> ├── settings_schema.json
> ├── locales/
>    ├── en.json
>    ├── fr.json
>    ├── ...
> ```

We haven't discussed that yet, however if it would be helpful to do so, I don't see why not. Could probably consider doing that through the CopyWebpackPlugin's transform as well.

harshal317:
> Since folders are looking to be introduced here would it be much extra effort to allow an additional layer of folders in the `sections` folder? For example:
> 
> ```
> ./sections
> ├── homepage/
>    ├── featured-collections/
>       ├── schema.json
>       ├── template.liquid
>    ├── hero/
>       ├── schema.json
>       ├── template.liquid
>    ├── ...
> ├── product/
>    ├── featured-articles/
>       ├── schema.json
>       ├── template.liquid
>    ├── variants/
>       ├── schema.json
>       ├── template.liquid
>    ├── ...
> ```

I believe we discussed having a deeper folder structure, and decided that it could be added as a feature moving forward.

chrisberthe:
>We haven't discussed that yet, however if it would be helpful to do so, I don't see why not.

It's definitely helpful, and it reflects how we're already approaching merchant-facing translations (this is how locales are currently formatted in our themes for both the `./sections` and `./config` folders).

dan-gamble:
Didn't think this would be in so quick. Can't wait for `0.16` 😍

PR #1001, opened 2019-03-05
Update Slate project status - Low maintenance

wizardlyhel:
Update README and git templates

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ![slate animated banner](https://user-images.githubusercontent.com/4837696/47506317-cbe22400-d83d-11e8-9867-1dc874943833.gif)
 
 ## Slate on Low Maintenance
-Due to project priority and resource constraints, this project is currently on low maintenance. We recognize that there are a lot of activities around this package, however, we unable to accommodate the maintenance this project requires.
+Due to project priority and resource constraints, this project is currently on low maintenance. We recognize that there are a lot of activities around this package. However, we are unable to accommodate the maintenance this project requires.
 
 ### What does this mean?
 Slate will not have new features or bug fixes during this period.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -2,6 +2,36 @@
 
 ![slate animated banner](https://user-images.githubusercontent.com/4837696/47506317-cbe22400-d83d-11e8-9867-1dc874943833.gif)
 
+## Slate on Low Maintenance
+Due to project priority and resource constraints, this project is currently on low maintenance. We recognize that there are a lot of activities around this package, however, we unable to accommodate the maintenance this project requires.
```
chrisberthe:
```suggestion
Due to project priority and resource constraints, this project is currently on low maintenance. We recognize that there are a lot of activities around this package. However, we are unable to accommodate the maintenance this project requires.
```

README.md
```diff
@@ -2,6 +2,36 @@
...
+- New feature implementation
+- Existing feature changes
+- Major bugs due to any tools version changes
+- Major bugs due to Shopify internal changes
+- Minor bugs
+- Will not evaluate and merge any community created PRs
+- Will not address any community reported issues except ones that implicate Shopify security
+
+### Should I continue using Slate?
+You may. As it is, Slate is still useful for themes development.
```
chrisberthe:
```suggestion
You may. As it is, Slate is still useful for Shopify theme development.
```

README.md
```diff
@@ -2,6 +2,36 @@
...
+
+### What happened? Why are you stopping? Will Shopify work on Slate in the future?
+Due to Shopify priorities and resource constraints, we have decided to put more emphasis on another project rather than Slate. We have plans to return back to Slate when our priorities shifts back to Slate. While evaluating the current state of Slate, there are two key developer workflows that we are not satisfied with:
```
chrisberthe:
```suggestion
Due to Shopify priorities and resource constraints, we have decided to put more emphasis on another projects. We have plans to return to Slate once our priorities change. While evaluating the current state of Slate, there are two key developer workflows that we are not satisfied with:
```

dan-gamble:
Is this likely to affect https://github.com/shopify/theme-scripts as well?

jonathanmoore:
Yikes! This is a pretty big setback, hopefully, other solutions and tools will emerge. I really liked where Slate was headed. 😞

wizardlyhel:
@dan-gamble theme-script is not affected. It is an individual utilities library that can be used independently without Slate.

@jonathanmoore Thank you for your feedback. This is not an easy decision either. With the increasing Slate usage week by week, as of late too, it is really hard for us to make this decision as well. However, it is also why we put it in low maintenance. We want to be better equipped when we do come back to Slate, so that we could tackle on the real, hard problems.

ruinyourfun:
Hire more talented developers as the ones you already have :) Anyway, hope you'll get back to it soon, this is a pretty ambitious & great project that I'll still use.

Issue #1004, opened 2019-03-06
Slate not watching header.liquid


masterShake:
npm start is triggering browser sync just fine and most of my files (templates, css, js) are being watched except section/header.liquid

Any idea how I can add this file to the watch list?

YourWishes:
I'm noticing this for all sections after beta 16.

Maurice-JB:
Same. Liquid files in sections\ folder not being watched. 
Due to changes from #1000?

dan-gamble:
Does it work if you use the new convention of `section/header/template.liquid` and `section/header/schema.json`?

dan-gamble:
Unfortunately, this is happening for everything inside sections including the new stuff.

I appreciate Slate has been put in low maintenance now but releasing a slightly broken release can hopefully be rectified before this low maintenance kicks into full effect @t-kelly @chrisberthe @wizardlyhel?

wizardlyhel:
Thank you @dan-gamble for the PR

dan-gamble:
Happy to help @wizardlyhel 😊

harshal317:
Fixed in #1005

PR #1005, opened 2019-03-07
Add the sections folder back to webpack context

dan-gamble:
Fix a small issue introduced by https://github.com/Shopify/slate/pull/1000.

With the new Slate Sections Plugin introduction, the `sections` folder got removed from Webpack's context. I believe it was previously added there by the CopyWebpackPlugin.

This PR adds it back into the context so making changes to the files causes the browser to reload as it used to.

https://github.com/Shopify/slate/issues/1004

### Checklist
For contributors:

### DIFF
--- a/packages/slate-sections-plugin/index.js
+++ b/packages/slate-sections-plugin/index.js
@@ -2,8 +2,6 @@ const fs = require('fs-extra');
 const path = require('path');
 const {ConcatSource, RawSource} = require('webpack-sources');
 const _ = require('lodash');
-const SlateConfig = require('@shopify/slate-config');
-const config = new SlateConfig(require('../slate-tools/slate-tools.schema'));
 
 const DEFAULT_GENERIC_TEMPLATE_NAME = 'template.liquid';
 const PLUGIN_NAME = 'Slate Sections Plugin';
@@ -16,10 +14,7 @@ module.exports = class sectionsPlugin {
   }
 
   apply(compiler) {
-    compiler.hooks.emit.tapPromise(
-      PLUGIN_NAME,
-      this.addLocales.bind(this),
-    );
+    compiler.hooks.emit.tapPromise(PLUGIN_NAME, this.addLocales.bind(this));
 
     compiler.hooks.afterEmit.tapPromise(
       PLUGIN_NAME,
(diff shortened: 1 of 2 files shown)


dan-gamble:
I signed the CLA but not quite sure how to re-run it.

harshal317:
Thanks a lot @dan-gamble, I made some styling changes so it would pass CI. I also got rid of the config import, as the sections path should be what we receive as the to option in the plugin, so figured it be best to just use that. I'll let @t-kelly take a look before publishing.

dan-gamble:
Awesome. I figured there would be some changes as I'm not that au fait with the Shopify coding standard.

treechime:
Thanks, @dan-gamble!

Can this be expected to be merged soon? The current version is very frustrating to develop with until this is fixed...

packages/slate-sections-plugin/index.js
```diff
@@ -13,9 +14,11 @@ module.exports = class sectionsPlugin {
   }
 
   apply(compiler) {
-    compiler.hooks.emit.tapPromise(
-      'Slate Sections Plugin',
-      this.addLocales.bind(this),
+    compiler.hooks.emit.tapPromise(PLUGIN_NAME, this.addLocales.bind(this));
+
+    compiler.hooks.afterEmit.tapPromise(
```
t-kelly: This doesn't need to be async, so just `compiler.hooks.emit.tap()` will work

packages/slate-sections-plugin/index.js
```diff
@@ -53,6 +56,12 @@ module.exports = class sectionsPlugin {
     );
   }
 
+  addSectionsToContext(compilation) {
+    compilation.contextDependencies.add(this.options.to);
+
+    return Promise.resolve({});
```
t-kelly: Once we use the synchronous `tap()`, we can remove this.

harshal317: I just added it to the other hook we tap, as it can be done there.

packages/slate-sections-plugin/index.js
harshal317: I just added it to the other hook we tap, as it can be done there.

harshal317:
> Thanks, @dan-gamble!
> 
> Can this be expected to be merged soon? The current version is very frustrating to develop with until this is fixed...

Sorry for the wait! Ideally we should get this merged today.

harshal317:
> Should we be adding the files from `this.options.from` instead of `this.options.to`?

That was a mistake, ended up just adding to the context when we tap the emit hook, and added making sure the sections folder is in the context to the tests. 🎩 locally worked.

treechime:
Perfect, thanks all 👍

PR #1010, opened 2019-03-20
Add @shopify/slate-translations plugin

chrisberthe:

We recently merged the `@shopify/slate-sections-plugin` into Slate. My current feels about the package: it focuses on one too many things.

1. Support for the section folder structure (this is perfect).
2. Support for separate locale files that then get merged into a schema upon build.

I feel like 2. should be encapsulated in its own package (e.g. `@shopify/slate-translations` (please let me know if this name sucks 🤦‍♂️, I had `slate-locales` at first 🤷‍♂️), that `@shopify/slate-sections-plugin` then imports. And this is what I've done in this PR.

Truth be told, I wouldn’t mind keeping the section plugin as is, but my concern lies when we’ll also need to support the possibility of adding a `locales/` folder in `src/config` - leveraging the same functionality as we currently have in `slate-sections-plugin` but for the `settings_schema.json` file. This will be tackled in a future PR.


dan-gamble:
This looks really awesome @chrisberthe! One question this raises with me is how this relates to https://github.com/Shopify/slate/pull/1001?

harshal317:
👋 I can help add the tests

chrisberthe:
@dan-gamble while Slate is currently in low maintenance mode, there are things we realize we need to fix to further our own internal theme development, like supporting the same translations folder structure for `src/config` as we do for sections.

While we no longer have many resources on this project, we'll progressively work on small incremental features that need taking-care-of, like this one. How this applies to the rest of Slate is detailed in the README - this PR (and the next that I will create for `src/config` translations) will most likely be the last PRs to be merged.

packages/slate-translations/index.js
```diff
@@ -0,0 +1,83 @@
...
+        if (typeof obj[key].t === 'string') {
+          obj[key] = await _getLocalizedValues(obj[key].t, localizedSchema);
+        } else if (typeof obj[key] === 'object') {
+          await traverse(obj[key]);
+        }
+      }),
+    );
+    return JSON.stringify(obj, null, 2);
+  };
+  const mainSchema = await fs.readJSON(mainSchemaPath, 'utf-8');
```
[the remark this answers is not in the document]
harshal317: I guess we have to read the locales anyways for combineLocales and that makes more sense than passing however many localized schema's as a parameter. So this is fine as well

packages/slate-sections-plugin/README.md
```diff
@@ -74,91 +74,3 @@ In order to have separate schema files, you must create a folder for the section
 ```
 
 This will create an `article-template.liquid` with the contents of the `template.liquid` file and with JSON wrapped in `{% schema %}` tags, which will be copied over to the `./dist` sections folder.
-
-## Translations
```
harshal317: If we're getting rid of this section we probably still want to specify how you can have a locales folder since right now it just shows having a schema.json and template.liquid in a folder, and a sections folder with just liquid files.

chrisberthe: Good call - I'll bring back this section but instead write a small description and point to the `@shopify/slate-translations` README.

packages/slate-translations/index.js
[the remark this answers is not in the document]
chrisberthe: As discussed in person, we'll this as is :).

dan-gamble:
Thank you for the explanation @chrisberthe! Very insightful.

chrisberthe:
Still wondering what the better name for this package is - have any thoughts on this @harshal317 @t-kelly? Perhaps being more descriptive is better, e.g. `@shopify/slate-schema-locales`.

harshal317:
> Still wondering what the better name for this package should be. Have any thoughts on this @harshal317 @t-kelly? Perhaps being more descriptive is better, e.g. `@shopify/slate-schema-locales` or `@shopify/slate-schema-translations`.

I like the name you have already `@shopify/slate-translations`, most intuitive name imo

packages/slate-translations/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@shopify/slate-translations",
+  "version": "1.0.0-beta.16",
```
t-kelly: Same as I mentioned above.

Issue #906, opened 2018-12-07
Slate v0 is not reloading browser after trying Slate V1


radmedov:
**### Problem**
I was trying to use Slate V1 on an existing project then I was back to Salte v0.
But the browsersync doesn't reload the browser anymore.

Any idea why?
P.S. I've removed mkcert, yarn and killed HSTS in chrome already.  

Thank you.

t-kelly:
@radmedov can you try viewing you Slate v0 project in a Chrome incognito window? Pretty sure this is because Slate v1 uses `https://localhost` and not `http`, which can cause Chrome to default to using `https` in the future.

radmedov:
@t-kelly It doesn't work in an incognito mode as well. It's also not working in any browser.
I'm getting in the Chrome console:

> This site does not have a valid SSL certificate! Without SSL, your site's and visitors' data is vulnerable to theft and tampering. Get a valid SSL certificate before releasing your website to the public. 

Firefox's console:

> This site makes use of a SHA-1 Certificate; it’s recommended you use certificates with signature algorithms that use hash functions stronger than SHA-1.

> 
Strict-Transport-Security: The connection to the site is untrustworthy, so the specified header was ignored.

Mkcert certificate was removed. The browser uses the certificate from the node-modules folder of the project.
I've tried to build a new slate v0 project with just slate default theme and it also doesn't reload browser. It seems like the browser wasn't using any certificate from node-modules before I've tried Slate v1. I just have no idea how to force browser not to use any certificate.

Thank you.

t-kelly:
@radmedov if you're in MacOS, try opening the Keychain Access app and deleting any cert referencing localhost?

![image](https://user-images.githubusercontent.com/4837696/49808542-ce83d480-fd2a-11e8-8ea9-8330e5056191.png)

radmedov:
@t-kelly, I did removed it immediately I got back to Slate v0. Sorry if I wasn’t clear at the beginning. I did remove all related to the Slate v1, re-install node.

lizsterine:
@radmedov Have you been able to fix your issue?  I just tried Slate 1 and now my Slate 0 isn't reloading either :( .

radmedov:
@lizsterine Nope, I gave up forcing it to work. I'm using Slate v1 now for developing from scratch.
[...]

PR #1046, opened 2019-07-18
Disable HSTS for dev server

stevekanter-arcticleaf:
Most other non-Slate 1.x development is via HTTP so by having HSTS on it really makes Shopify development tricky.

Most non-Shopify Slate 1.x development is done via HTTP. By leaving on Shopify's HSTS policy, it forces any localhost visits to happen via HTTPS. This simply removes that header from Shopify's proxied response so that HTTPS and HTTP can be used on localhost without issue.

https://github.com/Shopify/slate/issues/906

### Checklist
For contributors:

For maintainers:

### DIFF
--- a/packages/slate-tools/tools/dev-server/index.js
+++ b/packages/slate-tools/tools/dev-server/index.js
@@ -30,11 +30,11 @@ class DevServer {
           next();
         },
         proxyRes: [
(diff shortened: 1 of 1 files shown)


t-kelly:
Brilliant! 👍

PR #1058, opened 2019-09-21
Update Themekit version and themekit-node API

### LINKED ISSUE #1057: Slate-tools using themekit 0.6.12
### Problem

Slate-tools is using a very old version of themekit. This seems to be causing some sort of issue for us when uploading our /config/settings_data.json file with it.

### Replication steps

Run `yarn watch`
Update settings_data.json
In the background, the themekit command is run, but the file is not updated on Shopify
Running the same command with the current version of themekit 1.0.1 does update the file in shopify

### More Information

I'm running slate on windows - aware this isn't tested/officially supported but other than this issue it's been working fine (and themekit should actually work on windows anyway).

I'm happy to submit a PR for this, but I wanted to check if there was any reason for using an old version of themekit before doing so

JTimNolan:

fixes #1057 by updating themekit binary and themekit-node api calls to the latest version

### Checklist
For contributors:
- [N/A] I have [updated the docs](https://github.com/Shopify/slate/blob/master/CONTRIBUTING.md#documentation) to reflect these changes, if applicable.

For maintainers:
- [N/A] I have :tophat:'d these changes.


packages/slate-sync/index.js
```diff
@@ -78,9 +69,9 @@ function _generateIgnoreFlags() {
  * @return          Promise
  */
 async function deploy(cmd = '', files = []) {
-  if (!['upload', 'replace'].includes(cmd)) {
+  if (!['deploy', 'replace'].includes(cmd)) {
```
t-kelly: It looks like `replace` doesn't exist anymore? And has been replaced by `deploy`. And that `upload` has been replaced by `deploy --nodelete`?

JTimNolan: Yes, you're right. I'll update the the commands and documentation to reflect this when I get a chance

PR #1078, opened 2019-11-26
Disconnect Slate Analytics Package

### LINKED ISSUE #1075: Slate Analytics 404
### Problem
Shopify Slate Analytics has been disabled/broken/shut down which causes recurring 404s when using Slate. 

### Replication steps
- Install Slate v1 or use an existing Slate v1 project
     - If not using an existing project create a new project
- Run `watch` or `start` 

```bash
✔  Compiled successfully in 1.259s!
(node:50840) UnhandledPromiseRejectionWarning: Error: Request failed with status code 404
    at createError (/Users/username/slatedirectory/node_modules/axios/lib/core/createError.js:16:15)
    at settle (/Users/username/slatedirectory/node_modules/axios/lib/core/settle.js:18:12)
    at IncomingMessage.handleStreamEnd (/Users/username/slatedirectory/node_modules/axios/lib/adapters/http.js:202:11)
    at emitNone (events.js:111:20)
    at IncomingMessage.emit (events.js:208:7)
    at endReadableNT (_stream_readable.js:1064:12)
    at _combinedTickCallback (internal/process/next_tick.js:139:11)
    at process._tickCallback (internal/process/next_tick.js:181:9)
(node:50840) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 8)

★  You are editing files in theme 123456789 on the following store:

      https://myshopifystore.myshopify.com?preview_theme_id=123456789
```

### More Information

Because the analytics service has been disabled/broken/shut down, the 404 can be patched by disabling the Axios post in `node_modules/@shopify/slate-analytics/index.js`. As always this patch will only last until you install a project dependency or update project dependencies.

fix

curiouscrusher:
#### Removed Slate Analytics tracking due to service shutdown.
- Resolves #1075 
- Does not remove the `slate-analytics` package from the repo, only removes it as a dependency along with all of it's events in the various Slate tools. However if full removal is the direction that things should go I can adjust this accordingly. 

### Checklist
For contributors:
- [x] I have [updated the docs]
        I removed the sidebar link, should further explanation be added happy to add some content reflecting the removal.

### DIFF
--- a/docs/v1/sidebars.json
+++ b/docs/v1/sidebars.json
@@ -25,7 +25,6 @@
     ],
     "Reference": [
       "create-slate-theme",
-      "slate-analytics",
       "slate-configuration",
       "slate-cssvar-loader"
     ]
--- a/packages/create-slate-theme/create-slate-theme.js
+++ b/packages/create-slate-theme/create-slate-theme.js
@@ -1,7 +1,6 @@
 const hostedGitInfo = require('hosted-git-info');
 const validateProjectName = require('validate-npm-package-name');
 const env = require('@shopify/slate-env');
-const analytics = require('@shopify/slate-analytics');
 const fs = require('fs-extra');
 const path = require('path');
 const chalk = require('chalk');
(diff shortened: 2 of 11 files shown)


curiouscrusher:
Bumping for any updates

